PR #3, opened 2018-03-29
Thread mapping

ulysseB:
Changes the encoding of the mapping of thread dimensions to hardware dimensions. Reduces the memory pressure by a factor of 2x and the amount of deadends by a factor of 10x.

Needs  to:
- [x] Fix the computation of global parallelism
- [x] Individualise the waste computation
- [x] Make `bound_order` exploration work on mm
- [x] Assess the effect on bounds accuracy


ulysseB:
The amount of parallelism, computed in `model::local_info` is incorrect.

Issue #11, opened 2018-04-16
Remove Cargo.lock from the index


ulysseB:
Cargo.lock has been added to the index so we can pin rustlex dependencies. This forces us to compile with rustc version  `nightly-2018-04=02`. Once #4 is solved, we can remove `Cargo.lock` to compile with any rustc version.

ulysseB:
fixed by #16

Issue #4, opened 2018-03-29
Discard the Rustlex dependency


ulysseB:
Rustlex uses libsyntax and thus cannot run on stable. It must be ported each time the compiler internals change. The available solutions are:
- to use the Lalrpop tokenizer, but it is limited and cannot count lines for error reporting
- to use flex and generate a C tokenizer
- to use a parser combinator, such as nom, to create tokens.

ulysseB:
fixed by #16

PR #16, opened 2018-04-26
Stabilize Lexer from rust/lex as F/lex

adjivas:
Proposition of resolution for #4 and  #11's issue.
Thank you yet for the help @ulysseB.


Cargo.toml
```diff
@@ -86,6 +86,8 @@ cuda = []
 default = []
 format_exh = []
 mppa = []
+lex = []
```
ulysseB: Why add a `lex` feature ?

telamon-gen/build.rs
```diff
@@ -1,9 +1,32 @@
+extern crate cc;
 extern crate lalrpop;
 
+
 /// Adds a dependency to the build script.
 fn add_dependency(dep: &str) { println!("cargo:rerun-if-changed={}", dep); }
 
 fn main() {
+    // Compile the lexer.(`LEX="flex" cargo build --features "lex"`)
+    #[cfg(feature = "lex")]
```
ulysseB: Why do this only when the lex feature is activated ? Couldn't we always do this ?

telamon-gen/src/exh.l
```diff
@@ -0,0 +1,256 @@
...
+equal "="
+divide "/"
+
+choice_ident [a-z][a-z_0-9]*
+value_ident [A-Z][A-Z_0-9]*
+set_ident [A-Z][A-Za-z_0-9]*
+var \${alpha_num}+
+code \"[^\n\"]*\"
+
+/* https://github.com/a-haas/Compilation/blob/6fe46a65d64decdb648feba790bb7b1d2980ca0a/yacc/setparser.l */
```
ulysseB: What are these comment ? Should they stay here ?

telamon-gen/build.rs
adjivas: That not needed and will always require the f/lex dependency.
That why this feature is only here for regenerate the C file lexer if specified.

PR #17, opened 2018-05-02
Remove dependence on `drain_filter` feature from rust nightly

adjivas:
Remark:
* [L232](https://github.com/ulysseB/telamon/pull/17/commits/478505af28d03f0752fa1d21dda17db85bb75fa1#diff-effb07ac49bcd4c985d005f9ea90cde8R232) from [ir module](https://github.com/adjivas/telamon/blob/478505af28d03f0752fa1d21dda17db85bb75fa1/telamon-gen/src/ir/mod.rs) is uncovered by any test.


telamon-gen/src/ir/mod.rs
```diff
@@ -207,7 +209,8 @@ impl IrDesc {
         }
         // Reverse the set constraints when the set parameter is defined in the foralls.
         // TODO(cleanup): make the reversing code readable
-        for _ in set_constraints.drain_filter(|&mut (var, ref mut set)| {
+        // TODO(cleanup): reimplemente `drain_filter` when stable rust will be ready.
+        set_constraints.iter_mut().all(|&mut (var, ref mut set)| {
```
ulysseB: When you iterate on a structure without producing a value, prefere a `for` loop.

telamon-gen/src/ir/mod.rs
```diff
@@ -219,12 +222,20 @@ impl IrDesc {
...
                 *set = superset;
+            }
+            true
+        });
+        set_constraints.retain(|&(var, ref set)| {
+            if let Some(Variable::Forall(_)) = (&*set).arg() {
                 if let Variable::Arg(arg_id) = var {
                     let given_set = target.arguments().get(arg_id).1;
-                    given_set.is_subset_of_def(set)
+                    given_set.is_subset_of_def(set).not()
```
ulysseB: Prefer `!` than `.not()` as it is more idiomatic and does not require to import `std::ops::Not`.

PR #37, opened 2018-07-02
Multiline for Code  

adjivas:
Implementation of multiline for token code


telamon-gen/src/exh.l
```diff
@@ -120,9 +120,14 @@ code \"[^\n\"]*\"
...
-<LINE_DOC>\n      { yyextra.end.column = 0; BEGIN(INITIAL); }
-<LINE_DOC>[^\n]*  { yyextra.end.column += 3; return DOC; }
+{doc}                      { yyextra.end.column -= 3; BEGIN(LINE_DOC); }
+<LINE_DOC>\n               { yyextra.end.column = 0; BEGIN(INITIAL); }
+<LINE_DOC>[^\n]*           { yyextra.end.column += 3; return DOC; }
+
+\"                         { BEGIN(LINE_CODE); }
+<LINE_CODE>\"              { BEGIN(INITIAL); }
+<LINE_CODE>\\\n[ \t]*      { }
+<LINE_CODE>.               { return CODE; }
```
Elarnon: Is this matching a single character of the code segments then stitching them back? I suppose the parser is far from the bottleneck, but I think something like this instead should match larger chunks of text at once:

```
<LINE_CODE>[^\]+|\\ { return CODE; }
```

telamon-gen/src/lexer/mod.rs
```diff
@@ -70,6 +68,7 @@ pub type SpannedLexer<Tok, Pos, Err> = Result<(Pos, Tok, Pos), Err>;
 pub struct Lexer {
     scanner: YyScan,
     buffer: YyBufferState,
+    next: Option<SpannedLexer<Token, Position, LexicalError>>,
```
Elarnon: It is somewhat confusing to have the name of this field match the name of the `next` method. Maybe call it `next_token` (or `lookahead_token`) instead?

telamon-gen/src/lexer/mod.rs
```diff
@@ -126,108 +126,125 @@ impl Iterator for Lexer {
...
+                    },
+                    YyToken::Code => {
+                        let out = yyget_text(self.scanner);
+                        CStr::from_ptr(out)
+                             .to_str().ok()
+                             .and_then(|s: &str| {
+                                  let mut s: String = s.to_owned();
+                                  let mut e = extra.end;
+                                  loop {
+                                      let next = self.next();
```
Elarnon: Personal preference (feel free to ignore) but this is one case where I find a `match` more readable since we care about the value in both branches:

```
match self.next() {
    Some(Ok((_, Token::Code(ref code), end)) => {
        e = end;
        s.push_str(code);
    },
    next => {
        self.next = next;
        return Some(Ok((extra.beg, Token::Code(s), e)))
    }
}
```

telamon-gen/tests/lexer.rs
```diff
@@ -488,3 +481,19 @@ fn doc_mode() {
     // Line Comment MultiDoc's Token
     assert_eq!(Lexer::from(b"// comment \n // comment".to_vec()).collect::<Vec<_>>(), vec![]);
 }
+
+#[test]
+fn lexer_code_mode() {
```
Elarnon: This test is missing a case ensuring that the new feature (line splitted with \) works.

Elarnon: Thank you GitHub for parsing escape characters. I mean line splitted with "backslash followed by a newline".

telamon-gen/src/exh.l
adjivas: The idea is to match token code like:
```rust
"$code" -> "code"
"$code code" -> "code code"
"$code \\\n code" -> "code code"
"$code \\\n     code" -> "code code"
```

telamon-gen/src/lexer/mod.rs
adjivas: Your right, I will implement this pattern

telamon-gen/tests/lexer.rs
adjivas: Yes, I have to complete this test.
cc @github :cat:

telamon-gen/src/lexer/mod.rs
adjivas: I will

ulysseB:
I indeed seems that "$fun" " .exec" produces a single string. This is a good thing ! Please add a test case for it.

telamon-gen/src/lexer/mod.rs
ulysseB: I would tend to agree with Basile, but this is open to discussion. I tried to look at the behavior from clippy, but could not get the coresponding lint (`single_match_else`) to work. What does @adjivas think ?

adjivas: Isn't a pattern like `match -> stmt | _` but a `match -> stmt | stmt` where `stmt` is used in both cases so I think the recommendation of @Elarnon is best because that replaces a `let stmt, if else` pattern by a more readable match pattern.

telamon-gen/src/exh.l
adjivas: @Elarnon That need a bit of adaptation because `<LINE_CODE>[^\]+|\\` returns "_\"" for a input like "$_" but I can try to implement this.

Elarnon: Ah right, probably needs `<LINE_CODE>[^\\"]+|\\`

adjivas: That right, now is it solved.

telamon-gen/src/lexer/mod.rs
adjivas: Solved

telamon-gen/tests/lexer.rs
adjivas: Solved.

adjivas:
@ulysseB I have added the single code test case at [lexer.rs#L516](https://github.com/ulysseB/telamon/pull/37/commits/ebd8c1ef9f80160b55a60398027bd3d75909d8d3#diff-b6b62141a252308731b42af557fb7f66R520).

PR #43, opened 2018-07-05
Implement serde support for explorer::config::Config

Elarnon:
Previously explorer::config::Config serialization and deserialization
from the Settings.toml configuration file were implemented by
hand. This makes it tedious and error-prone to add new configuration
fields, especially when they require new structures or enums as will
probably be the case for future statistics and logging options.

This PR implements serialization and deserialization support for
explorer::config::Config and related inner
structures (explorer::config::SearchAlgorithm, etc.). The syntax of
the configuration file is identical as before, except that the
algorithm-related options are now under the [algorithm] key. This is
required because serde doesn't support de-serializing "flat" enums
with a default value for missing entries. The `generate_settings_file`
Python script is also updated to generate config files compatible with
the new format.

This means that sections such as:

```
algorithm = "bound_order"
```

must be replaced with:

```
[algorithm]
type = "bound_order"
```

and sections such as:

```
algorithm = "bandit"
monte_carlo = false
threshold = 7
```

should now be:

```
[algorithm]
type = "bandit"
monte_carlo = false
threshold = 7
```

Arguably, this provides a better structure and separation
of the config file.


tools/tune_bandit/generate_setting_files.py
```diff
@@ -5,61 +5,102 @@
...
+            print("Key {} should be a dict; got {}".format(
+                ".".join(path), value))
             return False
-    if key in ["delta", "stop_bound", "distance_to_best"]:
-        try:
-            t = float(val)
+        unknown = set(value.keys()) - set(spec.keys())
+        if unknown:
+            print("Keys {} are missing".format(
+                ", ".join(".".join(path + (key, )) for key in unknown)))
```
nicoTolly: Manque un return False

tools/tune_bandit/generate_setting_files.py
```diff
@@ -5,61 +5,102 @@
...
+}
 
-key_list = ["num_worker", "log_file",  "algorithm", "stop_bound",
-        "timeout", "delta", "distance_to_best", "monte_carlo","threshold",
-        "new_nodes_order", "old_nodes_order"]
-search_list = ["bound_order", "bandit"]
-new_order_list = ["api", "random", "bound", "weighted_random"]
-old_order_list = ["bound", "bandit", "weighted_random"]
-bool_list = ["true", "false"]
+def check(value, *, path=(), spec=spec):
```
nicoTolly: Décrire chaque argument

tools/tune_bandit/generate_setting_files.py
Elarnon: Changed to use exceptions instead to avoid this type of mistakes.

tools/tune_bandit/generate_setting_files.py
Elarnon: Done.

Elarnon:
@nicoTolly if you can do a quick second round to check everything is OK! In particular the "empty config file" fix we discussed offline.

PR #44, opened 2018-07-05
Count number of candidate evaluated

Elarnon:
In order to compare different algorithms, fixing the number of
candidates evaluated provided for a fairer comparison than using a
timeout because it allows comparisons to be done on different
machines (provided they use different GPUs), and relax the constraint
of having the machine unused during benchmarking. Since low execution
time is the final objective, we still track it, notably to ensure that
algorithms don't get too slow at producing candidates. However, even
in that case, it is arguably easier to improve an efficient algorithm
by making it faster rather than directly finding a fast and efficient
algorithm.

Actually limiting the number of evaluations performed is somewhat
tricky given the current architecture, so this currently implements an
approximate limit on the number of descents performed by the bandit
algorithm as a proxy instead, which is lower than the actual number of
evaluations.

### DIFF
--- a/src/explorer/bandit_arm.rs
+++ b/src/explorer/bandit_arm.rs
@@ -8,13 +8,15 @@ use explorer::store::Store;
 use itertools::Itertools;
 use std;
 use std::f64;
-use std::sync::{ Weak, Arc, RwLock};
+use std::sync::{Weak, Arc, RwLock};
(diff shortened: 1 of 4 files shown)


ulysseB:
I am not sure I understand why you can't look at he number evaluation. I think the monitor could count the number of evaluations it recieves and cut the search space when it expected number is reached. Or am I missing something ?

PR #48, opened 2018-07-10
Store gpu_data on a well-known path on the system

Elarnon:
Currently telamon assumes that a "data/cuda_gpus.json" file is
accessible from the directory where it is run. This is not optimal,
since we may want to run telamon from multiple directories, for
instance to load multiple settings file.

Instead, this makes it so that we use the "cuda_gpus.json" file from
~/.config/telamon/cuda_gpus.json and write a default "cuda_gpus.json"
configuration there if it doesn't exist. We use the `xdg` crate to
find the configuration directory in an appropriate way depending on
the system configuration, and load the default "data/cuda_gpus.json"
into the binary as a static string.

Note that in some cases (e.g. running tests) we may have race issues
if multiple instances of telamon are run at the same time and multiple
of them try to write the ~/.config/telamon/cuda_gpus.json. This would
only happen on the first run and shouldn't be a problem; however if it
ever becomes one we can easily change the behavior to use the bundled
cuda_gpus.json file and provide an additional utility to write the
proper configuration otherwise.

### DIFF
--- a/src/device/cuda/gpu.rs
+++ b/src/device/cuda/gpu.rs
@@ -1,4 +1,6 @@
 //! Describes CUDA-enabled GPUs.
+extern crate xdg;
+
 use codegen::Function;
 use device::{self, Device};
 use device::cuda::printer as p;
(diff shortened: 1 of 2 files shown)


src/device/cuda/gpu.rs
```diff
@@ -158,7 +160,26 @@ pub struct Gpu {
...
+                    let config_path = xdg_dirs
+                        .place_config_file("cuda_gpus.json")
+                        .expect("cannot create configuration directory");
+                    println!(
+                        "GPU configuration file not found; writing default configuration to {}",
+                        config_path.to_string_lossy());
+                    let mut config_file = unwrap!(File::create(config_path.clone()));
+                    unwrap!(
+                        write!(config_file, "{}", include_str!("../../../data/cuda_gpus.json")));
+                    config_path
```
ulysseB: Maybe you could return the file directly instead of returning the path, to avoid opening it twice. Not sure it is worth it though.

Elarnon: I actually explicitely wanted to open the file twice in order to ensure the final `config_file` is always opened in read-only mode. I will add a comment explaining this.

PR #53, opened 2018-07-12
Introduce bare-bones C and Python API

### LINKED ISSUE #49: Python bindings
We want to have Python bindings to Telamon in order to ease experimentation of statistical exploration approaches and integration with experiment reporting platforms such as mlflow.

Proposed approach is to use cbindgen to create a thin C API, then bundle it through milksnake to create a Python package. The initial goal is to expose a `search` function to drive Telamon's search from Python, as well as a couple helper to create kernels. Additional capabilities will be added as needed.

Elarnon:
This adds a new telamon-capi crate exporting a C API through a dynamic
library, and a Python package which uses said C API crate to expose a
Python API for Telamon. For the moment the API is very simple and only
allows optimising a MatMul kernel on either an x86 or cuda device, but
should evolve over time to incorporate additional features such as
driving the exploration from Python.

Fixes #49


telamon-py/telamon/__init__.py
```diff
@@ -0,0 +1,106 @@
+import contextlib
+import json
+from telamon._capi import ffi, lib
+
+class TelamonError(Exception):
+    """Base error class for Telamon errors."""
+
+# FIXME: The device stack should be a thread-local variable.
+_DEVICE_STACK = []
```
ulysseB: If I understand correctly, this is to make the `device` context reentrant ?

telamon-capi/src/lib.rs
```diff
@@ -0,0 +1,90 @@
+extern crate libc;
```
ulysseB: Adding at least a `//!` comment at the top of the file to explain waht is this library doing would be nice.

telamon-py/telamon/__init__.py
Elarnon: Exactly. In practice it is no different than a global "set_device" or some such call, but creates a cleaner API for defining the device to use only once without having to type it out for each experiment (when there are multiple experiments in the same Python script).

Elarnon:
I left the generated C header file in the tree in order to make it easier to use directly for potential future non-Python users of the C API (e.g. if the .so library was obtained by compiling on another system). It is probably a bit misguided however; I'll remove it.

PR #57, opened 2018-07-14
Expose the cuda module without the cuda feature

### LINKED ISSUE #52: Allow executors to fail
Currently  the available executors are defined based on feature flags and the complete function/type definitions are erased when the feature is disabled.

It would be useful to keep a thin API layer for interoperability even when the feature is disabled, and to make it so that the API can fail. This would allow:

 - Writing generic code which dynamically checks for a feature (e.g. using `cfg!(feature)` instead of `#[cfg(feature)]` and #[cfg(not(feature))]` pairs).
 - In a second time, allow features to be compiled in and fail gracefully (i.e. without panicking) if no hardware accelerator is found on the current machine.

ulysseB:
The PR partially fixes #52 . It makes most of the `device/cuda` module available without `--feature=cuda` and without requireing the CUDA driver to be available on the machine.

It works by creating a fake `Executor` (and a few other types linked to `Executor`) that cannot be created. It also add a `try_init` method to the `Executor` to dynamically test if the cuda feature is enable.


src/lib.rs
```diff
@@ -6,6 +6,8 @@ extern crate crossbeam;
 extern crate env_logger;
 extern crate getopts;
 extern crate errno;
+#[macro_use]
+extern crate failure;
```
Elarnon: This doesn't seem to be used?

Elarnon: Ah this is for the `#[fail]` annotation on `InitError`, my bad.

ulysseB: Yes. I just discovered the `failure` crate and it seems a good way to create error types. You can look at https://boats.gitlab.io/failure/ for more info.

ulysseB:
`@bors` doesn't works. You need to use `bors` instead. `@bors` is just for mozilla's repositories.

Elarnon:
Oh, ok. Kind of weird but well :)

PR #58, opened 2018-07-14
Automatically characterize GPUs

ulysseB:
We currently expect the user to have created a file contaning the characterization of the GPU. Instead we create this file automatically when it does not exists. We also use this opportunity to replace `rustc_serialize` by `serde`.

The cuda charactrize independent has been stripped down to only print the configuration on the screen. We may later totally remove it in favor of a Python API to call the characterization methods and manipulate the results of independent micro benchmarks.

This PR removes the outdate cache characterization binary, that was still maintaned to compile but not used for anything. It also changes the format of the JSON file to only store the description for a single GPU: except on very special cases, machines always have the same model on the different GPU slots and it makes the gestion of the file simpler.


src/bin/cuda_characterize.rs
```diff
@@ -0,0 +1,62 @@
...
+}
+
+fn main() {
+    env_logger::init();
+    let config = Config::parse();
+    let executor = cuda::Executor::init();
+    let gpu = cuda::characterize::characterize(&executor);
+
+    if config.write_to_file {
+        unimplemented!(); // FIXME
```
Elarnon: Should we leave this an `unimplemented!`? Do we still need this separate binary?

src/device/cuda/characterize/mod.rs
```diff
@@ -0,0 +1,78 @@
...
+fn create_table(parameters: &[&str], counters: &[cuda::PerfCounter]) -> Table<u64> {
+    let header = parameters.iter().map(|x| x.to_string())
+        .chain(counters.iter().map(|x| x.to_string())).collect_vec();
+    Table::new(header)
+}
+
+/// Returns the name of the configuration file.
+pub fn get_config_path() -> std::path::PathBuf {
+    let xdg_dirs = unwrap!(xdg::BaseDirectories::with_prefix("telamon"));
+    xdg_dirs.find_config_file("cuda_gpus.json").unwrap_or_else(|| {
```
Elarnon: I think this should only be `xdg_dirs.place_config_file("cuda_gpus.json")`, which will return `~/.config/telamon/cuda_gpus.json`. `xdg_dirs.find_config_file` could return system-wide files e.g. in `/usr/share/telamon/cuda_gpus.json` (although those will probably never exist in practice) on which the user doesn't have write permissions.

src/device/cuda/characterize/mod.rs
```diff
@@ -0,0 +1,78 @@
...
+            warn!("GPU description file is already in use, waiting");
+            file.lock_exclusive()
+        },
+        _ => Err(err)
+    }));
+    unwrap!(file.seek(std::io::SeekFrom::Start(0)));
+    let gpu = serde_json::from_reader(&file).map_err(Error::Parser)
+        .and_then(|gpu: cuda::Gpu| {
+            let name = executor.device_name();
+            if gpu.name == name { Ok(gpu) } else { Err(Error::WrongGpu(name)) }
```
Elarnon: If I am reading this correctly, it means the `cuda_gpus.json` config file can only have description for a single GPU? I thought we had support for a list of different GPUs.

src/device/cuda/characterize/mod.rs
```diff
@@ -0,0 +1,78 @@
...
+        }).unwrap_or_else(|err| {
+            warn!("{}. Runing characterization.", err);
+            let gpu = characterize(executor);
+            // FIXME: print the GPU
+            gpu
+        });
+    unwrap!(file.unlock());
```
Elarnon: I don't know the file locking APIs, but is there a way to do the locking/unlocking using a contextual object triggering unlock at end of scope?

src/device/cuda/characterize/mod.rs
```diff
@@ -0,0 +1,78 @@
...
+#[derive(Debug, Fail)]
+enum Error {
+    #[fail(display="could not parse GPU description: {}", _0)]
+    Parser(serde_json::Error),
+    #[fail(display="found description for the wrong GPU: {}", _0)]
+    WrongGpu(String),
+}
+
+/// Retrieve the description of the GPU from the description file. Updates it if needed.
+pub fn get_gpu_desc(executor: &cuda::Executor) -> cuda::Gpu {
```
Elarnon: Is this function actually updating the config file if the appropriate GPU was not found? I could not find that piece of code.

Elarnon:
(This was "changes requested" but should probably have been "simple comment", oops)

src/device/cuda/characterize/mod.rs
ulysseB: Outdated since I removed the file locking.

src/device/cuda/characterize/mod.rs
ulysseB: I changed to remove the support for multiple GPUs. This way we always overwrite the file if we need to re-run the characterization, which makes things simpler.

If we want a pre-computed characterization file, then we will need to add support for multiple GPUs.

src/device/cuda/characterize/mod.rs
ulysseB: Should be OK in the newest commit.

ulysseB:
I think we should **not** distribute a default `cuda_gpu.json`. Indeed, while the characterization can be slow, it takes at most ~10 minutes, which is still reasonable compared to the usual search time. We just need to make sure it is not wipped out on test/benchmarking machines (can we do that with Jenkins ?).

Maintining the file for a range of GPUs will be cumbersome as we will need to collect and merge data from sevral sources each time the GPU description is changed.

src/device/cuda/characterize/mod.rs
ulysseB: Fixed

src/device/cuda/characterize/mod.rs
```diff
@@ -0,0 +1,72 @@
...
+    let config_path = get_config_path();
+    lazy_static! {
```
Elarnon: I didn't know you could have static variables inside a function in Rust. Does this have the same semantics as in e.g. C?

src/bin/cuda_characterize.rs
ulysseB: Now implemented

Elarnon:
That makes sense I guess. Yeah it should work with Jenkins since we are storing stuff in ~/.config which is not inside the workspace and thus not erased across builds. It is a bit hackish however because it won't work very well when we have builds using different characterization format at the same time :) But we can figure out a solution at that time.

src/device/cuda/characterize/mod.rs
ulysseB: Indeed.

PR #59, opened 2018-07-16
Benchmark Decents in the Search Tree

### LINKED ISSUE #7: Create benchmarks to assess telamon-gen output performance
For some kernels, the code generated by telamon-gen is limiting the speed of the exploration. We need to precisely asses the performance of telamon-gen, both for Telamon and for dedicated benchmarks.

ulysseB:
This PR adds a benchmark to measure the time it takes to descend in the search tree. The benchmarks are based on the criterion framework and measure the performance of a descent when generating code for GPUs.

We measure both decents with and without saving the encountered candidates to measure the impact of copies on the runtime. These benchmarks meant to measure the whole descent, individual benchmarks for specific telamon-gen features will later be needed but not part of this PR.

Fixes #7

### DIFF
--- a/benches/descent.rs
+++ b/benches/descent.rs
@@ -1,0 +1,149 @@
+//! Benchmarks how fast it is to decend in the search tree.
+#[macro_use]
+extern crate criterion;
+extern crate env_logger;
+extern crate telamon;
+extern crate rand;
+#[macro_use]
+extern crate lazy_static;
+
+use criterion::Criterion;
+use telamon::{ir, device, explorer, helper};
+use telamon::explorer::choice::ActionEx;
+use telamon::search_space::*;
+use rand::Rng;
+
+/// Configure the bencher.
+fn config_criterion() -> Criterion {
+    Criterion::default()
+        .sample_size(50)
+        .configure_from_args()
+}
+
+lazy_static! {
+    /// A fake GPU description, used only to know which candidates are valid.
+    static ref DEVICE: device::cuda::Gpu = device::cuda::Gpu::dummy();
+
+    static ref MM_SIGNATURE: MMSig = MMSig::signature();
+    static ref MM: SearchSpace<'static> = MM_SIGNATURE.build_body();
+}
+
+/// Stores the signature and the external arrays IDs for matrix-matrix multiplication.
+struct MMSig {
+    signature: ir::Signature,
+    a: ir::mem::Id,
+    b: ir::mem::Id,
+    c: ir::mem::Id,
+}
+
+impl MMSig {
+    fn signature() -> Self {
+        let mut signature = ir::Signature::new("mm".to_string());
+        signature.add_scalar("m".to_string(), ir::Type::I(32));
+        signature.add_scalar("n".to_string(), ir::Type::I(32));
+        signature.add_scalar("k".to_string(), ir::Type::I(32));
+        let a = signature.add_array("a".to_string());
+        let b = signature.add_array("b".to_string());
+        let c = signature.add_array("c".to_string());
+        MMSig { signature, a, b, c }
+    }
+
+    fn build_body(&self) -> SearchSpace {
+        const DATA_TYPE: ir::Type = ir::Type::F(32);
+        let mut builder = helper::Builder::new(&self.signature, &*DEVICE);
+        let m_size = builder.param_size("m");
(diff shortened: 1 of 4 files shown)


benches/descent.rs
```diff
@@ -0,0 +1,149 @@
...
+
+/// Benchmarks full descents in the search tree, with a copy at each level.
+fn mm_descent_copy(c: &mut Criterion) {
+    let _ = env_logger::try_init();
+    c.bench_function("mm descent with copy", |b| b.iter(|| {
+        let mut spaces = vec![];
+        let mut space = MM.clone();
+        while let Some(mut choice) = {
+            let choice = explorer::choice::list(&space).next();
+            choice
```
Elarnon: Same as above.

benches/descent.rs
```diff
@@ -0,0 +1,149 @@
...
+    }
+}
+
+/// Benchmarks full descents in the search tree.
+fn mm_descent(c: &mut Criterion) {
+    let _ = env_logger::try_init();
+    c.bench_function("mm descent", |b| b.iter(|| {
+        let mut space = MM.clone();
+        while let Some(mut choice) = {
+            let choice = explorer::choice::list(&space).next();
```
Elarnon: Why this `let choice = ...; choice` instead of `while let Some(mut choice) = explorer::choice::list(&space).next() {`?

benches/descent.rs
```diff
@@ -0,0 +1,149 @@
...
+            choice
+        } {
+           let id = rand::thread_rng().gen_range(0, choice.len());
+           let res = match choice.swap_remove(id) {
+                ActionEx::TileSizes(..) => panic!(),
```
Elarnon: Oh, does ".." means "any number of arguments; I don't care about those" (i.e. a variadic variant of "_")? Neat if so!

benches/descent.rs
```diff
@@ -0,0 +1,149 @@
...
+                ActionEx::Action(action) => space.apply_decisions(vec![action]),
+                ActionEx::LowerLayout { mem, ref st_dims, ref ld_dims } =>
+                    space.lower_layout(mem, st_dims.clone(), ld_dims.clone()),
+           };
+           if res.is_err() { return; }
+        }
+    }));
+}
+
+/// Benchmarks full descents in the search tree, with a copy at each level.
```
Elarnon: Why do we care about benchmarking the descent with an extra copy?

benches/descent.rs
ulysseB: Yes.

benches/descent.rs
ulysseB: Some wired limitation of the borrow checker. If I don't do that, it thinks space is borrowed in the body of the loop.

benches/descent.rs
ulysseB: see below.

benches/descent.rs
ulysseB: This is what happens in practice: when you create multiple children for a node, you need to copy the search space before applying each decision. I wanted to make sur the copy overhead was small enough.

benches/descent.rs
Elarnon: Oh right. Lexical lifetime issue? Maybe leave a comment there to explain what is going on.

benches/descent.rs
Elarnon: That makes sense. Is it possible to make parameterized benches (I don't know if this exists) and use an if statement there to avoid duplicating the code?

adjivas:
A Criterion output is generated at `target/criterion/`
We can move it into `target/doc` for add this output to the gh-pages.

Criterion has a [gnuplot feature](https://github.com/japaric/criterion.rs#features), so, we can add it to the travis manifest like:
```
addons:
  packages:
      # Criterion's dependency Quickstart § https://github.com/japaric/criterion.rs/blob/master/README.md#quickstart
      - gnuplot
```

Creterion can comparure between two `cargo bench` the difference, with a integration continue we can compare the two last commits according to this solution [https://github.com/japaric/criterion.rs/blob/master/book/src/faq.md#how-should-i-run-criterionrs-benchmarks-in-a-ci-pipeline](https://github.com/japaric/criterion.rs/blob/master/book/src/faq.md#how-should-i-run-criterionrs-benchmarks-in-a-ci-pipeline) by example

PR #61, opened 2018-07-16
Improve the performance of constraint propagation

ulysseB:
Constraint propagation is the main bottleneck appart from evaluation on the targeted device. This PR improves performance by avoiding calling multiple times the same propagator, improving performance by 30%.

In details, it ensure filters are only called on the lower triangular part of symmetric decisons since the upper triangular part is fowarded to the lower triangular part.


Elarnon:
In practice this is only going to matter for `DimKind` choices, right?

bors r+

PR #62, opened 2018-07-16
Add ability to configure matmul tilings from the Python API

Elarnon:
This allows manually overriding the tiling configuration for the MatMul
kernel through parameters in MatMulP and exposes that configuration
through the C and Python API.

In addition, a new Python example script is included which runs an
optimal candidate search from Python with a couple options (e.g. kernel
and device selection) which can be used as a starting point for further
experiments.

### DIFF
--- a/kernels/src/linalg.rs
+++ b/kernels/src/linalg.rs
@@ -260,7 +260,7 @@ pub struct MatMul<'a, S: Scalar> {
     c: Tensor<'a, S>,
 }
 
-#[derive(Clone, Copy)]
+#[derive(Clone)]
 pub struct MatMulP {
     pub m: i32,
     pub n: i32,
@@ -269,6 +269,9 @@ pub struct MatMulP {
     pub transpose_a: bool,
     pub transpose_b: bool,
     pub generic: bool,
+    pub m_tiling: Option<Vec<u32>>,
+    pub n_tiling: Option<Vec<u32>>,
+    pub k_tiling: Option<Vec<u32>>,
 }
 
 impl MatMulP {
@@ -279,6 +282,9 @@ impl MatMulP {
             transpose_a: false,
             transpose_b: false,
             generic: true,
+            m_tiling: None,
+            n_tiling: None,
+            k_tiling: None,
         }
     }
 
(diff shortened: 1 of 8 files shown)


kernels/src/linalg.rs
```diff
@@ -331,9 +337,21 @@ impl<'a, S: Scalar> Kernel<'a> for MatMul<'a, S> {
...
+        let m_tiles = if let Some(ref tiles) = self.params.m_tiling {
+            vec![tiles.clone()]
+        } else {
+            ::generate_tile_sizes(self.params.m as u32, &[64, 8])
+        };
+        let n_tiles = if let Some(ref tiles) = self.params.k_tiling {
+            vec![tiles.clone()]
+        } else {
+            ::generate_tile_sizes(self.params.n as u32, &[64, 8])
+        };
```
ulysseB: I tend to prefer `unwrap_or_else` to `if let` when returning a value. But this is discutable.
```
let k_tiles = self.params.k_tiling.clone().map(Vec::new).unwrap_or_else(|| {
    ::generate_tile_sizes(self.params.k as u32, &[64])
})
```

telamon-capi/src/lib.rs
```diff
@@ -15,6 +17,12 @@ use telamon::device::x86;
 use telamon::explorer::config::Config;
 pub use telamon_kernels::{linalg, Kernel};
 
+/// Initializes the logger.
+#[no_mangle]
+pub extern "C" fn env_logger_try_init() {
+    env_logger::try_init();
```
ulysseB: You should use `let _ = env_logger::try_init();` instead to avoid warnings because of the unused `Result`.

Elarnon: Ah, I was wondering what was the appropriate construct to use to silence those warnings! Will use this.

kernels/src/linalg.rs
Elarnon: That looks nicer; I'll change the code.

ulysseB:
bors retry

ulysseB:
My bad, I gave you the wrong code. Its `.map(|x| vec![x])` instead of `.map(Vec::new)`.

Elarnon:
bors retry

PR #63, opened 2018-07-16
WIP: Expose data-flow information in Telamon

### LINKED ISSUE #41: Exopose values and variables in the IR
Data flow is currently implicit in our representation. This causes some problems:
- we can't measure the register pressure
- it is hard to measure the size of memory blocks. This leads to a O(n^3) search space representation.
- we don't know when to introduce broadcasts
- reductions are hard to manipulate and we cannot represent parallel reductions
- we do not know when to privatise shared memory

Bellow is the modifications necessary to implement the new value system:
- [ ] Create the value structure and exopose the set of values in 'choices.exh'
- [ ] Add a new operand kind to read a value
- [ ] Allow to store the result of a computation in a value
- [ ] Handle the last operator
- [ ] Handle the fby operator
- [ ] Handle dimensions mapping
- [ ] Update the builder to use values and remove inst ansd reduce operands.

ulysseB:
Telamon's IR currently represents values as instruction inputs with `ir::Operand`.
Unfortunately, this does not allows us to manipulate decisions about operands and to
reason about the data flow in the program. This PR introduces a new type `ir::Value`
that represents a value stored in a variable. It lays the foundations for #41.

**This PR is a work in progress, do not merge**

### DIFF
--- a/src/ir/function.rs
+++ b/src/ir/function.rs
@@ -39,6 +39,7 @@ pub struct Function<'a> {
     mem_blocks: mem::BlockMap<'a>,
     layouts_to_lower: Vec<ir::mem::InternalId>,
     induction_vars: Vec<ir::InductionVar<'a>>,
+    values: Vec<ir::Value>,
 }
 
 impl<'a> Function<'a> {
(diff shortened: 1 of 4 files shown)


ulysseB:
Covered by #113

PR #65, opened 2018-07-17
Return errors instead of panicking when creating an IR instance.

ulysseB:
This PR introduces new error types and uses them by returning `Results<_, ir::Error>` instead of panicking when adding objects to the IR. This will be useful to create IR instances from an external scripting language without exiting the shell each time an error is raised.


ulysseB:
bors retry

PR #66, opened 2018-07-18
Benchmark the amount of memory used.

ulysseB:
This PR adds a benchmark to measure the amount of memory used by a single search space. It relies on jemalloc statistics. It reuses logic from the `descent` benchmark. The common code is thus moved to `benches/common/mod.rs`.

### DIFF
--- a/benches/memory.rs
+++ b/benches/memory.rs
@@ -1,0 +1,35 @@
+//! Measures the memory used by an IR instance.
+extern crate env_logger;
+extern crate itertools;
+extern crate jemalloc_ctl;
+#[macro_use]
+extern crate lazy_static;
+#[macro_use]
+extern crate telamon_utils;
+
+mod common;
+
+use itertools::Itertools;
+
+/// Reads the amount of resident memory.
+fn resident_memory() -> usize {
+    // many statistics are cached and only updated when the epoch is advanced.
+    unwrap!(jemalloc_ctl::epoch());
+    unwrap!(jemalloc_ctl::stats::resident())
+}
+
+const NUM_DESCENTS: usize = 1000;
+
+fn main() {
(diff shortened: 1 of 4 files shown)


benches/memory.rs
```diff
@@ -0,0 +1,28 @@
...
+use itertools::Itertools;
+
+/// Reads the amount of resident memory.
+fn resident_memory() -> usize {
+    // many statistics are cached and only updated when the epoch is advanced.
+    unwrap!(jemalloc_ctl::epoch());
+    unwrap!(jemalloc_ctl::stats::resident())
+}
+
+const NUM_DESCENTS: usize = 1000;
```
Elarnon: We don't actually do any descent here I think? Should we?

ulysseB: We should. The problem is their is sharing between the `SearchSpace` instances in the descent. Depending on the decisions being take, the sharing can be either big or small. The amount of memory used in a descent also depends on the number of deadends encountered. I am thus unsure on how to proceed to get meaningful numbers. If you have any suggestions ...

ulysseB: Maybe we could just apply a fixed list of actions, but this is not future-proof for when we change `choices.exh`.

Elarnon: Why can't we do the same descents as in the timing benchmark? Is it because it would be too noisy?

Maybe what we could do is have a thread that periodically samples memory usage and report statistics during the actual cuda_search benchmark (although we may want to have a simpler benchmark than that one...).

ulysseB: It would be noisy, but also make less sense. What we would need is to measure the memory used by a whole search tree after N iterations. I will do that, but later and only if I need it.

bors r+

PR #68, opened 2018-07-18
Compute the number of deadends encountered during search

Elarnon:


src/explorer/parallel_list.rs
```diff
@@ -30,6 +30,10 @@ impl<'a> Store<'a> for ParallelCandidateList<'a> {
             else {return None;}
         }
     }
+
+    fn print_stats(&self) {
+        // We don't have any stats to print.
+    }
 }
```
ulysseB: You can add a default implementation for the method in the `explorer::Store` trait.

src/explorer/bandit_arm.rs
```diff
@@ -73,20 +89,22 @@ impl<'a, 'b> Tree<'a, 'b> {
...
-                            Ok(candidate) => {
-                                let res = local_selection::descend(
-                                    self.config.new_nodes_order, context, candidate, cut);
-                                return res.map(|c| (c, path));
+                        if let Some((idx, maybe_candidate)) = res {
+                            path.0.push((Arc::downgrade(&node), idx));
+                            match maybe_candidate {
+                                Ok(candidate) => {
+                                    let res = local_selection::descend(
+                                        self.config.new_nodes_order, context, candidate, cut);
```
ulysseB: A dead-end can also occure here, when `local_selection::descend` returns `None`. If that happens, it means the random descent encountered a dead-end. In that case, we don't cleanup because the dead-end is not in the expanded part of the search tree, but we still need to account for it in the counter.

src/explorer/bandit_arm.rs
```diff
@@ -64,6 +79,7 @@ impl<'a, 'b> Tree<'a, 'b> {
         loop {
             match std::mem::replace(&mut state, DescendState::DeadEnd) {
                 DescendState::DeadEnd => {
+                    self.stats.num_deadends.fetch_add(1, Ordering::Relaxed);
```
ulysseB: You can actually count the number of deadends directly from the `Tree::explore` method. Indeed the `descend` method returns `None` if and only if a deadend is reached. This accounts both for dead-ends you captured here and the ones you did not capture.

Elarnon:
Deadends are now counted directly from the `explore` method.

PR #69, opened 2018-07-18
Add a Symmetric Difference method to `VecSet`.

ulysseB:
This PR adds a method to perform a symmetric difference on sets backed by vectors. It uses unsafe code to perform the operation in-place.

The unsafe code is required because with have temporary holes in the vectors we process: the data inside the holes is either droped or moved and we so we don't want it to be dropped a second time if a panic occurs (for example in the comparison function, potentially defined by the user, for the data type).

The trick is to set the lenght of arrays at 0 during the processing and set it back at the real lengh at the end. This way, if an unwinding occurs, it will not drop the content of the array and will only leak memory instead of accessing freed memory.

### DIFF
--- a/telamon-utils/src/vec_set.rs
+++ b/telamon-utils/src/vec_set.rs
@@ -372,10 +372,10 @@ mod tests {
     }
 
     #[test]
-    fn symmetric_difference() {
+    fn relative_difference() {
         let v0 = VecSet::new(vec![0, 2, 3, 5]);
         let v1 = VecSet::new(vec![0, 1, 2, 4, 6]);
-        let (d0, d1) = v0.symmetric_difference(v1);
+        let (d0, d1) = v0.relative_difference(v1);
         assert_eq!(d0, VecSet::new(vec![3, 5]));
         assert_eq!(d1, VecSet::new(vec![1, 4, 6]))
     }


telamon-utils/src/vec_set.rs
```diff
@@ -37,6 +37,62 @@ where T: Ord
...
+    /// Returns `self\other` and `other\self`.
+    pub fn symmetric_difference(mut self, mut other: Self) -> (Self, Self) {
+        let lhs_old_len = self.data.len();
+        let rhs_old_len = other.data.len();
+        unsafe {
+            // Make sure the moved elements are not drop twice if an exception occurs.
+            self.data.set_len(0);
+            other.data.set_len(0);
+            // Iterate simultaneously on both vectors to remove duplicate elements.
+            let (mut lhs_idx, mut lhs_del, mut rhs_idx, mut rhs_del) = (0, 0, 0, 0);
```
Elarnon: Why are those not of type `usize`?

telamon-utils/src/vec_set.rs
```diff
@@ -37,6 +37,62 @@ where T: Ord
...
+            while lhs_idx < lhs_old_len && rhs_idx < rhs_old_len {
+                let lhs = self.data.as_mut_ptr().offset(lhs_idx as isize);
```
Elarnon: Why not `let lhs: *mut T = &mut self.data[lhs_idx]`? Is it to avoid bound checks?

telamon-utils/src/vec_set.rs
```diff
@@ -37,6 +37,62 @@ where T: Ord
...
+                let rhs = other.data.as_mut_ptr().offset(rhs_idx as isize);
+                match (*lhs).cmp(&*rhs) {
+                    std::cmp::Ordering::Less => {
+                        if lhs_del > 0 {
+                            let lhs_dst = self.data.as_mut_ptr().offset((lhs_idx - lhs_del) as isize);
+                            std::ptr::copy_nonoverlapping(lhs, lhs_dst, 1);
```
Elarnon: Why are we using `copy_nonoverlapping` instead of `ptr::write(lhs_dst, ptr::read(lhs))`?

telamon-utils/src/vec_set.rs
```diff
@@ -37,6 +37,62 @@ where T: Ord
...
+                    },
+                    std::cmp::Ordering::Greater => {
+                        if rhs_del > 0 {
+                            let rhs_dst = other.data.as_mut_ptr().offset((rhs_idx - rhs_del) as isize);
+                            std::ptr::copy_nonoverlapping(rhs, rhs_dst, 1);
+                        }
+                        rhs_idx += 1;
+                    },
+                    std::cmp::Ordering::Equal => {
+                        std::mem::drop(&*lhs);
```
Elarnon: I am not sure it matters that much, but since we are explicitely avoiding copies and other bounds checks all over the place -- why not `ptr::drop_in_place(lhs)`?

telamon-utils/src/vec_set.rs
```diff
@@ -37,6 +37,62 @@ where T: Ord
...
+                        lhs_idx += 1;
+                        lhs_del += 1;
+                        rhs_idx += 1;
+                        rhs_del += 1;
+                    }
+                }
+            }
+            // Complete vectors that are not yet explored.
+            if lhs_idx < lhs_old_len && lhs_del > 0 {
+                let lhs_src = self.data.as_mut_ptr().offset(lhs_idx as isize);
```
Elarnon: This should be `self.data.as_ptr` since we are not writing to `lhs_src`.

telamon-utils/src/vec_set.rs
Elarnon: I guess it is.

telamon-utils/src/vec_set.rs
ulysseB: Fixed

telamon-utils/src/vec_set.rs
ulysseB: Indeed, fixed.

telamon-utils/src/vec_set.rs
ulysseB: Since the lenght of arrays has been set to `0`, bound checks will fail so we must bypass them.

telamon-utils/src/vec_set.rs
ulysseB: They are of type `usize`.

telamon-utils/src/vec_set.rs
ulysseB: `copy_nonoverlapping` can call `memcopy`, which is heavily optimized. `ptr::write(ptr::read(..))` might be replaced by a `memcopy` by the compiler but their is no grarantee.

telamon-utils/src/vec_set.rs
```diff
@@ -37,6 +37,62 @@ where T: Ord
         }
     }
 
+    /// Returns `self\other` and `other\self`.
+    pub fn symmetric_difference(mut self, mut other: Self) -> (Self, Self) {
```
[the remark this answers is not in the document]
ulysseB: I always thought this operation was called "Symmetric Difference". Thank you for pointing out it was something else.

ulysseB:
bors retry

telamon-utils/src/vec_set.rs
Elarnon: The more you know. On a first approach it seemed weird to use an array API for a single element :)

PR #71, opened 2018-07-18
WIP: revamp numeric sets

### LINKED ISSUE #46: Find a more efficient representation for `NumericSet`.
NumericSet allocates arrays of 16 `u32` on the stacks. Instead we could either:
- only keep a reference to the universe and tick available posibilities in a bitset. The universe should, in this case, be stored in the sigature.
- pass the universe as argument each time we make a computation

ulysseB:
Fixes #46.

**This PR is a work in progress, do not merge.**


ulysseB:
closing in favor of smaller PRs

PR #72, opened 2018-07-18
Include

adjivas:
The keyword include described by this grammar rule:
```lex
"include".*\".*\"
```
* The `.*` can be replaced by a limitation.
* The relative paths of exh's include depends of the cargo environment.
```rust
let filename: &str = concat!(env!("CARGO_MANIFEST_DIR"), "/tests/enum_foo.exh");
let include = format!("include \"{}\"", filename);
```
Maybe we can consider a default relative folder path to includes the exh header files than a path who will depend of the working directory of Cargo.


adjivas:
The include is now restricted for `"include"{whitespace}*\".*\"`

adjivas:
The include and sub-include is tested.
There miss a include guard, this was notified at [lexer_include_guard](https://github.com/adjivas/telamon/blob/include/telamon-gen/tests/lexer.rs#L571) test case.

telamon-gen/src/ast/choice/enumeration.rs
```diff
@@ -18,20 +18,22 @@ impl EnumDef {
             match stmt {
                 EnumStatement::Value(Spanned { beg, end, data: name }, ..) => {
                     hash.insert(name.to_owned(), Spanned {
-                        beg: *beg, end: *end, data: ()
+                        beg: beg.to_owned(), end: end.to_owned(),
```
ulysseB: This pattern seems to occure a lot in your code. Maybe you could create a method to copy the positions with a new data.
```
fn with_data<T>(&self, data: T)  -> Spanned<T>
```

adjivas: Solved

ulysseB: Thanks, I will look at this.

adjivas:
@ulysseB , I have added this explicite [test](https://github.com/ulysseB/telamon/pull/72/commits/448d3a2a774b1022b3adb0febd157d8bb39afb63#diff-b6b62141a252308731b42af557fb7f66R570).

PR #73, opened 2018-07-18
Actually fix use-after-free in Python API

Elarnon:
As it turns out, cffi actually handles properly calls taking a <cdata>
as argument and decrements references afterwards, which makes the
previous fix useless. However, storing a <cdata> pointer into another
<cdata> *doesn't* count as a reference -- hence, the `Tiling` object was
still allocated, but pointed to an integer buffer which had been freed,
and *that* was the source of the use-after-free.

This changes the API slightly so that we use a wrapper class instead
that we ensure outlives the C call.

### DIFF
--- a/telamon-capi/src/lib.rs
+++ b/telamon-capi/src/lib.rs
@@ -53,27 +53,11 @@ impl KernelParameters {
     }
 }
 
-
-#[repr(C)]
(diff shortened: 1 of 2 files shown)


Elarnon:
After offline discussion -- let's use numpy buffers instead.

PR #74, opened 2018-07-18
Extend NumericSet

ulysseB:
This PR extends the range of values `NumericSet` can store and adds new methods to it.

### DIFF
--- a/telamon-gen/cc_tests/Cargo.toml
+++ b/telamon-gen/cc_tests/Cargo.toml
@@ -14,6 +14,7 @@ path = "../"
 env_logger = "0.5.9"
 itertools = "0.7.8"
 log = "0.4.1"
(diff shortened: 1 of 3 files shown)


ulysseB:
It is used to store the size of loops, for example 2^25 on axpy. Another approach would be to separate the cases were the size is above 2^16 to handle them as if the size was unknown. The later might be  good idea since big dimensions will behave as dimensions with a big dynamic size anyway since they cannot be `UNROLL`, `THREAD` or `VECTOR`. However it might not be future-proof if we want to use `NumericSet` somewhere else.

A better approach would be to store the possible values elsewhere and only store a bitfield where each bit corresponds to a possible value in the domain. This would save place without relying on `u16`.

ulysseB:
bors retry

PR #79, opened 2018-07-19
Actually prevent use-after-free in Python API

Elarnon:
Storing a <cdata> pointer, as we were doing, into another <cdata>
structure *does not* count as a reference; the Python <cdata> object
must be kept alive for the pointed-to C memory to be valid.

Instead, we will now fully allocate the memory from Rust and use a
temporary NumPy array to copy the data from. This means we are currently
performing three copies from a Python list to a Rust Vec: one copy to
the NumPy array, then a second copy from the NumPy array to the kernel
parameters structure. Users can avoid the first copy by passing in an
existing NumPy array instead of a list, in which case we read from that
array's buffer directly. The second copy is required because we cannot
transfer ownership from Python to Rust due to different allocators.

### DIFF
--- a/telamon-capi/src/lib.rs
+++ b/telamon-capi/src/lib.rs
@@ -9,7 +9,7 @@ extern crate libc;
 extern crate telamon;
 extern crate telamon_kernels;
 
-use libc::{c_char, c_int, c_uint, size_t};
+use libc::{c_char, c_int, c_uint, size_t, uint32_t};
 use telamon::device;
 #[cfg(feature = "cuda")]
 use telamon::device::cuda;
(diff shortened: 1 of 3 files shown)


telamon-capi/src/lib.rs
```diff
@@ -46,39 +46,28 @@ impl KernelParameters {
...
-    data: *mut c_uint,
-}
-
-unsafe fn tiling_as_vec(tiling: *const Tiling) -> Option<Vec<u32>> {
-    if tiling.is_null() {
+/// Helper function to create a Rust vector from a C array (pointer
+/// and size) without transfering ownership (it performs a
+/// copy). Returns None when data is null.
+unsafe fn c_vec<T: Clone>(data: *const T, len: usize) -> Option<Vec<T>> {
+    if data.is_null() {
```
ulysseB: Why not return an empty vector instead ?

Elarnon: An empty vector can be created with a zero length and a dangling pointer (this is the behavior one would get currently when creating a NumPy array from an empty list in Python). We need a way to encode "None" which is the value we use to tell the underlying Rust API to call `generate_tiles`.

ulysseB:
I am not sure if you are finished with this. I will let you r+ it if you are.

Elarnon:
I am; I was waiting on your r+ to have proper bors metadata for once. Oh well :)

bors r+

PR #80, opened 2018-07-19
Expose the set of dimensions with a static size.

ulysseB:
This PR improves the encoding of the search space by exposing dimensions with a statically known size in a dedicated set. It allows us to restrict the scope of many constraints and of the `thread_mapping` choice. It also allows to access the size of such dimensions without providing a default value.

Later, it will allow us to expose  a `size` choice for such dimensions.

This PR greatly improves the exploration performance thanks to the reduced number of constraints to propagate: 1.8x factor on the descent time.


ulysseB:
bors retry

PR #82, opened 2018-07-24
Ensure telamon-gen output is deterministic

Elarnon:
We want telamon-gen output to be deterministic on a given .exh file
because we want to be able to assume field and enum order when
serializing to compact binary formats. Currently telamon-gen is
deterministic (thanks to using the non-random HashMaps from
telamon-utils) up to the printer, where we use a topological sort from
the `pathfinding` crate to ensure counters are serialized after their
dependencies. Unfortunately, the topological sort in the `pathfinding`
crate is non-deterministic due to its use of regular Rust HashMaps.

This patch replaces the topological sort from the `pathfinding` crate
with our own topological sort implementation,
`stable_topological_sort`, which we guarantee is not only
non-deterministic but is also stable: the order between elements which
don't have (transitive) dependencies to each other is preserved.

In addition, it replaces the use of some HashMaps with IndexMaps from
the `indexmap` crate, which are hash tables ensuring that iteration
order is identical to insertion order until any value is removed. This
allows to help propagate the stability guarantee to the order in which
choices are defined in the `.exh` file, giving some control on the
generated code to the user.

### DIFF
--- a/telamon-gen/src/print/mod.rs
+++ b/telamon-gen/src/print/mod.rs
@@ -377,7 +377,7 @@ where
 
     // At each iteration of this loop we maintain the invariant that
     // the priority queue contains exactly the nodes for which all
-    // predecessors are in sorted, but have not been sorted
+    // predecessors are already sorted, but have not been sorted
     // themselves. Using a priority queue ensures that we are always
     // processing nodes in the order they appear in the initial
     // `nodes` slice.
@@ -418,7 +418,7 @@ where
     }
 }
 
-/// Note that the code in the `stable_topological_sort_tests` omdule
+/// Note that the code in the `stable_topological_sort_tests` module
 /// should be in an Examples sections in the `stable_topological_sort`
 /// function, but we can't run rustdoc tests on private functions for
 /// some reason.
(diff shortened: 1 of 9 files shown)


telamon-gen/src/print/mod.rs
```diff
@@ -240,10 +246,265 @@ pub fn print(ir_desc: &ir::IrDesc) -> String {
...
+    let mut queue = BinaryHeap::with_capacity(nodes.len());
+    for data in node_data.iter_mut() {
+        if data.as_mut().unwrap().num_predecessors == 0 {
+            queue.push(data.take().unwrap())
+        }
+    }
+
+    // At each iteration of this loop we maintain the invariant that
+    // the priority queue contains exactly the nodes for which all
+    // predecessors are in sorted, but have not been sorted
```
ulysseB: are in sorted -> are sorted

telamon-gen/src/print/mod.rs
```diff
@@ -240,10 +246,265 @@ pub fn print(ir_desc: &ir::IrDesc) -> String {
...
+                cycles.push(nodes[index].clone())
+            }
+        }
+        Err((sorted, cycles))
+    } else {
+        Ok(sorted)
+    }
+}
+
+/// Note that the code in the `stable_topological_sort_tests` omdule
```
ulysseB: omdule -> module

Elarnon:
bors r=ulysseB

PR #84, opened 2018-07-24
Benchmark the variance of measures on the device

ulysseB:
This PR exposes a problem with the measures on the device. It compares the execution time obtained after multiple evaluation on the GPU. It simulates both the case where the GPU is the limiting factor and the case where the CPU is the limiting factor.

It shows that when the CPU is limiting, measures are noisy and innacurate, sometime by a factor of more than 2.

### DIFF
--- a/kernels/benches/cuda_variance.rs
+++ b/kernels/benches/cuda_variance.rs
@@ -1,5 +1,6 @@
 //! Benchmarks the variance of measures on the GPU.
 extern crate env_logger;
+extern crate futures;
 extern crate itertools;
 #[macro_use]
 extern crate log;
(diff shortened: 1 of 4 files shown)


kernels/benches/cuda_variance.rs
```diff
@@ -0,0 +1,116 @@
...
+use utils::*;
+
+/// The number of candidates to try.
+const NUM_TESTS: usize = 20;
+/// The number of time each candidate should be evaluated, in cpu-bound mode.
+const NUM_SLOW_SAMPLES: usize = 20;
+/// The number of time each candidate should be evaluated, in gpu-bound mode.
+const NUM_FAST_SAMPLES: usize = 50;
+/// Maximal legal difference between evaluation times.
+const MAX_DIFF: f64 = 0.01;
```
Elarnon: Isn't this very high since we target execution times of around `0.001`? Would it makes sense to use a relative threshold instead (e.g. `1%` of the min of two execution times)?

Elarnon: Ah looks like this is already used as a relative difference. Please clarify this in the doc.

ulysseB:
bors r=elarnon

Issue #51, opened 2018-07-11
Dump an event log


Elarnon:
In order to better devise new possible ways of exploring the search space, it is useful to be able to get access to statistics on what happens during the search. For instance we may want to collect statistics on how discriminant some choices were. Currently we need to perform those statistics online while the search is running, but it would be useful to be able to dump an event log from which we could replay the whole search procedure and compute any new statistics we didn't compute at the time of the run.

Replaying the search should be significantly faster than re-starting it, especially since we won't run on the accelerator device nor the (expensive) propagation steps during the monte-carlo descents, and would allow to gather statistics after the fact that we didn't compute (maybe because we didn't know we wanted to compute them at the time) during the run. Proposed implementation is to dump that information into a protobuf stream. One issue is we need to generate .protos from telamon-gen, and we need a marker to ensure that we can only load the protos if the search space has not changed.

PR #86, opened 2018-07-25
Dump the list of candidates evaluated.

Elarnon:
This introduce the base architecture towards resolution of #51.

The event log is dumped as a .tfrecord file containing a sequence of
event values encoded with bincode. The use of tfrecord as a container
format means we can easily change the underlying serialization
protocol if needed.

The event log depends on the candidate store implementation, and
different candidate store can have different event types and logging
strategies. Currently, only the candidates evaluated by the bandit
algorithm are logged (with their evaluation time); a next step will be
to also log information about dead ends so that we are able to replay
the exploration after the fact for statistical exploration.

Currently, candidates are stored by serializing the list of actions
taken from the root of the tree to generate them, which requires
out-of-band transmission of the initial kernel definition. A next step
will be to serialize the "pristine" domain and IR definition so that
we don't need that out-of-band information to re-build the actual
candidates from the information in the event log anymore.

Note that bincode serialization depends on the order in which fields
are defined, and as such we can expect that event logs generated with
a given version of telamon may not work properly with another version,
if the Action representation has changed (not when using the same
version, thanks to #82). This is desirable, since any statistics
computed after the representation has changed would be bogus
anyways. However, it is a bit brittle currently because there is no
version number and we should consider emitting a special initial event
indicating such metadata.


Elarnon:
bors retry

Elarnon:
bors retry

Elarnon:
bors retry

PR #87, opened 2018-07-25
Spanned name for Var Definition

adjivas:
Short PR to prepare EnumDef check cases of issue #5 .


telamon-gen/src/ast/context.rs
```diff
@@ -362,8 +394,8 @@ impl TypingContext {
         // Type-check the base.
         let kind = body.kind;
         let all_var_defs = untyped_vars.iter().chain(&body.iter_vars).cloned().collect();
-        let vars = untyped_vars.into_iter().map(|def| {
-            (def.name.clone(), var_map.decl_argument(&self.ir_desc, def))
+        let vars = untyped_vars.iter().map(|def| {
+            (def.name.clone(), var_map.decl_argument(&self.ir_desc, def.clone()))
```
ulysseB: Why do you need to add a copy ?

adjivas: Specifically, it's a regression, I will have to restore it.

adjivas:
[constrain.rs](https://github.com/ulysseB/telamon/blob/dfe2979bcc0d4933d9316dcc3714a8e6626f7476/telamon-gen/src/ast/constrain.rs) and [constraint.rs](https://github.com/ulysseB/telamon/blob/dfe2979bcc0d4933d9316dcc3714a8e6626f7476/telamon-gen/src/ast/constraint.rs) was same, that was a error of copy, that why I have removed this doublon.

PR #88, opened 2018-07-25
Telamon C API for Cuda

ulysseB:
This PR extends telamon-capi to create and manipulate Cuda devices and contexts. It also creates two wrapped types `Context` and `Device` to hold pointers to trait objects of the same types in Rust.

As a side effect, the former `Device` enum has been renamed into `DeviceId`. I guess we must update the python API to reflect this change, but I have no idea where.

### DIFF
--- a/telamon-capi/src/cuda.rs
+++ b/telamon-capi/src/cuda.rs
@@ -1,0 +1,39 @@
+//! C API wrappers for manipulating Telamon CUDA context through FFI.
+use env_logger;
+use std;
+use telamon::device::cuda;
+
+/// Opaque type that bundles a `telamon::device::cuda::Context` and its executor.
+pub struct CudaContext {
+    executor: cuda::Executor,
+    context: std::mem::ManuallyDrop<cuda::Context<'static>>,
+}
+
+/// Returns a pointer to a CUDA context. The CUDA context is not the same than
+/// `telamon::device::cuda::Context` and should only be used with the C API functions.
+#[no_mangle]
+pub unsafe extern "C" fn telamon_cuda_create_context() -> *mut CudaContext {
+    let _ = env_logger::try_init();
+    let mut c_context = Box::new(CudaContext {
+        executor: cuda::Executor::init(),
+        context: std::mem::uninitialized(),
+    });
+    let context = std::mem::transmute(cuda::Context::new(&c_context.executor));
+    std::ptr::write(&mut c_context.context, std::mem::ManuallyDrop::new(context));
+    Box::into_raw(c_context)
+}
+
+/// Destroys a CUDA context.
+#[no_mangle]
+pub unsafe extern "C" fn telamon_cuda_destroy_context(ctx: *mut CudaContext) {
+    let mut ctx = Box::from_raw(ctx);
+    std::mem::ManuallyDrop::drop(&mut ctx.context);
+}
+
+#[test]
+fn create_destroy_context() {
+    unsafe {
+        let ctx = telamon_cuda_create_context();
+        telamon_cuda_destroy_context(ctx);
+    }
+}
--- a/telamon-capi/src/cuda.rs
+++ b/telamon-capi/src/cuda.rs
@@ -1,39 +1,104 @@
 //! C API wrappers for manipulating Telamon CUDA context through FFI.
+use Context;
 use env_logger;
+use libc;
 use std;
-use telamon::device::cuda;
+use telamon::device::{cuda, ArgMap};
+use telamon::ir;
 
-/// Opaque type that bundles a `telamon::device::cuda::Context` and its executor.
-pub struct CudaContext {
+/// Opaque type that contains the mapping of kernel parameters to actual values.
+pub struct CudaEnvironment {
     executor: cuda::Executor,
     context: std::mem::ManuallyDrop<cuda::Context<'static>>,
+    context_ref: Context,
 }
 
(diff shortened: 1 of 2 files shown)


telamon-capi/src/lib.rs
```diff
@@ -9,14 +9,22 @@ extern crate libc;
...
 use telamon::device::x86;
 use telamon::explorer::config::Config;
 pub use telamon_kernels::{linalg, Kernel};
 
+/// Description of the evaluation context. In particular, in contains the mapping between
+/// argument names and argument values.
+pub struct Context(*const device::Context);
+
+/// Description of the targeted device.
+pub struct Device(*const device::Device);
```
Elarnon: Are we missing either constructors or `#[repr(C)]` for those two types? They currently get exposed as opaque structs but we don't have ways of interacting with those opaque struct.

telamon-capi/src/cuda.rs
```diff
@@ -0,0 +1,116 @@
...
+use env_logger;
+use libc;
+use std;
+use telamon::device::{cuda, ArgMap};
+use telamon::ir;
+
+/// Opaque type that contains the mapping of kernel parameters to actual values.
+pub struct CudaEnvironment {
+    executor: cuda::Executor,
+    context: std::mem::ManuallyDrop<cuda::Context<'static>>,
```
Elarnon: Can you add a comment explaining why we need `ManuallyDrop` here? Is it in order to ensure that we don't drop the `executor` before dropping the `context`?

telamon-capi/src/cuda.rs
```diff
@@ -0,0 +1,116 @@
...
+    context_ref: Context,
+    device_ref: Device,
+}
+
+/// Returns a pointer to a CUDA environement. The caller is responsible for deallocating
+/// the pointer by calling `telamon_cuda_destroy_environment`.
+#[no_mangle]
+pub unsafe extern "C" fn telamon_cuda_create_environment() -> *mut CudaEnvironment {
```
Elarnon: Please call this `telamon_cuda_environment_create`. Having the `_create` suffix (rather than buried in the middle) makes it easier to see at a glance that this allocates and initializes memory (same thing for `telamon_cuda_destroy_context` which should be `telamon_cuda_environment_destroy`).

By the way, I have used `_new`/`_free` suffixes for the kernel Python API (this is the convention from GTK I think). I am fine with switching to `_create` / `_destroy` instead, but we should stay consistent (so either rename those to use `_new` and `_free`, or rename the `kernel` functions in `lib.rs` to use `_create` and `_destroy` instead).

telamon-capi/src/lib.rs
Elarnon: Actually those are used by the cuda module. Why do we need this extra layer of indirection exactly?

telamon-capi/src/cuda.rs
Elarnon: (Also do we actually need to erase the lifetime here?)

telamon-capi/src/cuda.rs
```diff
@@ -0,0 +1,116 @@
...
+    let _ = env_logger::try_init();
+    let mut env = Box::new(CudaEnvironment {
+        executor: cuda::Executor::init(),
+        context: std::mem::uninitialized(),
```
Elarnon: Those `mem::uninitialized` and `mem::transmute` deserve a short comment justifying that we need to have references in `context` to the `executor`, hence we need to first allocate the memory for the `executor` by placing it in the box before initializing the `context`.

telamon-capi/src/lib.rs
ulysseB: `device::Context` and `device::Device` are traits. Pointers to traits are actually fat points, with a type similar to https://doc.rust-lang.org/std/raw/struct.TraitObject.html:
```
pub struct TraitObject {
    pub data: *mut (),
    pub vtable: *mut (),
}
```

We have two alternatives to expose fat pointers in C:
- either we create a `#[repr(C)]` struct similar to `TraitObject`
- or we hide the fat pointer in an opaque struct and give pointers to the opaque struct to C users
The first solution avoids the additional level of indirection but exposes internal rust details to the user so I went with the second.

telamon-capi/src/cuda.rs
ulysseB: I will adopt your conventions.

telamon-capi/src/cuda.rs
ulysseB: Comment added. The lifetime needs to be erased because we do not know the lifetime of `executor`: it depends on when will the struct will be moved. Any lifetime parameter of `context` would need to be a parameter of the struct, but we cannot guarantee such a parameter will have the lifetime of the struct

A solution for that is available in nightly https://github.com/rust-lang/rfcs/blob/master/text/2349-pin.md.

Elarnon: Right, I was thinking we could return a `&'a mut CudaEnvironment<'a>` but it doesn't really make sense since we won't use those functins from rust anyways.

telamon-capi/src/lib.rs
ulysseB: I agree with the explicit `dyn`. I just tested and it is stable. Maybe we can wait for clippy to be stable (in 2 rust releases IIRC) so we can add it to the lints. Overwise, we are going to have `dyn` in half the places and not the others, which might leads to some confusion.

ulysseB:
bors r=elarnon

PR #89, opened 2018-07-25
Enumeration check

adjivas:
Check case for #5 .

* [x] undefined_field_antisymmetric: A antisymmetric should refers a defined fields from same enum.


adjivas:
This PR is ready for a review.

telamon-gen/src/ast/choice/enumeration.rs
```diff
@@ -10,64 +10,180 @@ pub struct EnumDef {
...
                     }
                 },
+            }
+
+        }
+        Ok(())
+    }
+
+    /// This checks that there isn't any doublon in parameter list.
+    fn check_redefinition_parameter(&self) -> Result<(), TypeError> {
```
ulysseB: This should be a method common to all `ChoiceDef`.

telamon-gen/src/ast/mod.rs
```diff
@@ -67,9 +67,11 @@ pub enum TypeError {
     /// Undefinition of set, enum or field.
     Undefined(Spanned<String>),
     /// Unvalid arguments of a symmetric enum.
-    BadSymmetricArg(Vec<VarDef>),
+    BadSymmetricArg(Spanned<String>, Vec<VarDef>),
```
ulysseB: Add a comment to explain what is the `Spanned<String>` or use named fields.

telamon-gen/src/ast/choice/enumeration.rs
adjivas: That solved by the [PR 94](https://github.com/ulysseB/telamon/pull/94)?

telamon-gen/src/ast/mod.rs
adjivas: Solved with a Alias _SpannedName_ and _Name_

telamon-gen/src/ast/choice/enumeration.rs
ulysseB: Ok, I will look for this in #94

telamon-gen/src/ast/mod.rs
ulysseB: Still not clear for me. I think the alias adds complexity but does not explains things. I would rather have a comment instead.

ulysseB:
I think you misunderstood me. The problem was not in the structs (where the field name was already providing the info) but the the definition of the `TypeError` enum, specifically the fields of  `BadSymmetricArg(Spanned<String>, Vec<VarDef>),`. Either document the or given them a name.

adjivas:
@ulysseB yes right, normally that fixed.
Ready for review again.

telamon-gen/src/ast/error.rs
```diff
@@ -0,0 +1,48 @@
...
+#[derive(Debug, PartialEq)]
+pub enum TypeError {
+    /// Redefinition of a name and hint.
+    Redefinition { object_kind: Spanned<Hint>, object_name: Spanned<String> },
+    /// Undefinition of set, enum or field.
+    Undefined { object_name: Spanned<String> },
+    /// Unvalid arguments of a symmetric enum.
+    BadSymmetricArg { object_name: Spanned<String>, object_variables: Vec<VarDef> },
+    /// Missing
+    /// Append when the Set's object has a missing field.
```
ulysseB: typo: Append -> Happens, idem for `Conflict`

PR #91, opened 2018-07-30
Improve the stability of measures on the GPU.

### LINKED ISSUE #85: Stabilize measures on the GPU when the CPU is limiting
When the CPU is limiting and the GPU underused, measures are innacurate: #84 shows a difference that can reach a factor of 2 between the measured time when the CPU is limiting compared to when the GPU is limiting.

Ideally, the fix should not impact performance: more evaluations are needed only in the case where the CPU was limiting. Bellow are multiple ideas to have stable measures:
- launch dummy kernel when the GPU is not working to keep it busy
- find another approach to pin the GPU clock
  - not available on ficus
- use events to detect when the measured and actual time do not match
  - some incorrect time may not be detected if the two are too close
- lauch each kernels multiple times.
  - exit early if the perf is too bad

ulysseB:
This PR fixes #85.

It makes the cuda context run each kernel 20 times if the first evaluation is within 3 times the execution time of the best candidate. The `cuda_variance` benchmark shows this is enough to have a delta inferior at 1% on 50 calls to the evaluation method.

Better accuracy at a much smaller cost is achievable by exiting after a few evaluation if the execution time is not good enough. This will be done in a future PR.

### DIFF
--- a/src/device/cuda/context.rs
+++ b/src/device/cuda/context.rs
@@ -6,6 +6,7 @@ use device::cuda::api::{self, Argument};
 use device::cuda::kernel::Thunk;
 use explorer;
 use ir;
+use itertools::*;
 use std;
 use std::f64;
 use std::sync::{atomic, mpsc, Arc};
@@ -16,6 +17,15 @@ use device::context::AsyncCallback;
 const EVAL_BUFFER_SIZE: usize = 100;
 // TODO(perf): enable optimizations when possible
 const JIT_OPT_LEVEL: usize = 2;
+
+/// Candidates with a runtime above `SKIP_THRESHOLD * cut` are skipped after the first
+/// evaluation.
+const SKIP_THRESHOLD: f64 = 3.;
+// FIXME: tune values + add a second threshold after a few iterations
+/// Number of evaluations of perform on each candidate.
+const NUM_EVALS: usize = 20;
+/// Number of outlier evaluations to discard.
+const NUM_OUTLIERS: usize = 4;
 
 /// A CUDA evaluation context.
 pub struct Context<'a> {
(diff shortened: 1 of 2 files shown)


src/device/cuda/context.rs
```diff
@@ -60,6 +70,40 @@ impl<'a> Context<'a> {
...
+        let t0 = self.ticks_to_ns(unwrap!(thunk.execute()));
+        if mode.skip_bad_candidates() && t0 * SKIP_THRESHOLD >= bound {
+            info!("candidate skipped after its first evaluation");
+            return Ok(t0);
+        }
+        // TODO(cc_perf): becomes the limiting factor after a few hours. We should stop
+        // earlier and make tests to know when (for example, measure the MAX delta between
+        // min and median with N outliers).
+        let runtimes = (0..NUM_EVALS).map(|_| thunk.execute());
+        let runtimes_by_value = process_results(runtimes, |iter| iter.sorted())?;
```
Elarnon: Is `process_results` defined somewhere?

src/device/cuda/context.rs
```diff
@@ -60,6 +70,40 @@ impl<'a> Context<'a> {
...
+        }
+        let t0 = self.ticks_to_ns(unwrap!(thunk.execute()));
+        if mode.skip_bad_candidates() && t0 * SKIP_THRESHOLD >= bound {
+            info!("candidate skipped after its first evaluation");
+            return Ok(t0);
+        }
+        // TODO(cc_perf): becomes the limiting factor after a few hours. We should stop
+        // earlier and make tests to know when (for example, measure the MAX delta between
+        // min and median with N outliers).
+        let runtimes = (0..NUM_EVALS).map(|_| thunk.execute());
```
Elarnon: Does this mean the kernel will get added `NUM_EVALS` times to the execution queue after the previous one has finished (i.e. they will not be consecutive)? Not that I'm sure whether this is a good or a bad thing...

ulysseB: No. `thunk.executes()` directly executes the kernel and returns its execution time. This code is called by the code that consumes the execution queue.

src/device/cuda/context.rs
ulysseB: It is imported by `use itertools::*;`. I will make the import explicit.

src/device/cuda/context.rs
Elarnon: Ah I see

PR #95, opened 2018-07-30
Spanned Key from Set

adjivas:
Key from Set wasn't Spanned, this PR span it as described by the [set::redefinition::field](https://github.com/adjivas/telamon/blob/1f359e8723506193cb11d3c51867205fc432799a/telamon-gen/tests/typing/set.rs#L61) check.


ulysseB:
Please add a description of what this doesm

PR #97, opened 2018-08-01
Pre-allocate IDs for lowering

Elarnon:
Currently telamon is dynamically creating new instruction and dimension
IDs when performing a lowering, which means that:

 - Depending on previous choices, the same ID can be used for different
   choices semantically.
 - The total number of possible choices is not known until all the
   lowerings have been performed (or not).

Both of those properties make it awkward to build statistical models on
top of the information present in the domain. This PR addresses those
issues by pre-allocating IDs for the instruction and dimensions that
will need to be created when a dimension map is lowered.

That information is stored on the `DimMap` instance associated with a
lowerable operand (currently only `Inst`) when the scope is global, and
is annotated on the creation of the DomainStore. This means that an
`ir::Function` can exist in an "invalid" non-annotated state prior to be
stored in a `SearchSpace`. However, that is not an issue, because the
function is not supposed to be used outside of the search space, and
additional instructions (except for lowering) are not supposed to be
added once the function is in the search space.

### DIFF
--- a/src/ir/dim_map.rs
+++ b/src/ir/dim_map.rs
@@ -5,11 +5,12 @@ use utils::*;
 
 /// Represents a mapping between dimenions.
 #[derive(Clone, Debug)]
-// TODO(cleanup): once merge is handled exclusively from the domain, we can use a `Vec`
-// instead.
+// TODO(cleanup): once merge is handled exclusively from the domain, we can use
+// a `Vec` instead.
 pub struct DimMap {
     map: LinkedList<(ir::DimId, ir::DimId)>,
 }
+
 
 // TODO(cleanup): Send should be derived for LinkedList.
 unsafe impl Send for DimMap {}
@@ -18,42 +19,61 @@ unsafe impl Sync for DimMap {}
 impl DimMap {
     /// Create a new `DimMap`.
     pub fn new<IT>(dims: IT) -> Self
-            where IT: IntoIterator<Item=(ir::DimId, ir::DimId)> {
-        DimMap { map: dims.into_iter().collect() }
+    where
+        IT: IntoIterator<Item = (ir::DimId, ir::DimId)>,
+    {
+        DimMap {
+            map: dims.into_iter().collect(),
+        }
     }
 
     /// Returns an empty `DimMap`.
-    pub fn empty() -> DimMap { DimMap { map: LinkedList::new() } }
+    pub fn empty() -> Self {
+        DimMap {
+            map: LinkedList::new(),
+        }
+    }
 
-    /// Renames a basic block into an other. Indicates if some mapping were removed.
+    /// Renames a basic block into an other. Indicates if some mapping were
+    /// removed.
     pub fn merge_dims(&mut self, lhs: ir::DimId, rhs: ir::DimId) -> bool {
-        self.filter(|&mut pair| pair == (lhs, rhs) || pair == (rhs, lhs)).count() > 0
+        self.filter(|&mut pair| pair == (lhs, rhs) || pair == (rhs, lhs))
+            .count() > 0
     }
 
     /// Iterates over the DimMap.
-    pub fn iter(&self) -> linked_list::Iter<(ir::DimId, ir::DimId)> { self.map.iter() }
+    pub fn iter(&self) -> linked_list::Iter<(ir::DimId, ir::DimId)> {
+        self.map.iter()
+    }
 
     /// Filters the DimMap.
     pub fn filter<F>(&mut self, f: F) -> FilterList<(ir::DimId, ir::DimId), F>
-            where F: FnMut(&mut (ir::DimId, ir::DimId)) -> bool {
+    where
+        F: FnMut(&mut (ir::DimId, ir::DimId)) -> bool,
+    {
         filter_list(&mut self.map, f)
     }
 
     /// Returns true if the `DimMap` is empty.
-    pub fn is_empty(&self) -> bool { self.map.is_empty() }
+    pub fn is_empty(&self) -> bool {
+        self.map.is_empty()
+    }
 }
 
 impl IntoIterator for DimMap {
     type Item = (ir::DimId, ir::DimId);
     type IntoIter = linked_list::IntoIter<(ir::DimId, ir::DimId)>;
 
-    fn into_iter(self) -> Self::IntoIter { self.map.into_iter() }
+    fn into_iter(self) -> Self::IntoIter {
+        self.map.into_iter()
+    }
 }
-
 
 impl<'a> IntoIterator for &'a DimMap {
     type Item = &'a (ir::DimId, ir::DimId);
     type IntoIter = linked_list::Iter<'a, (ir::DimId, ir::DimId)>;
 
-    fn into_iter(self) -> Self::IntoIter { self.map.iter() }
+    fn into_iter(self) -> Self::IntoIter {
+        self.map.iter()
+    }
 }
--- a/src/ir/function.rs
+++ b/src/ir/function.rs
@@ -1,7 +1,7 @@
 //! Provides a representation of functions.
 use device::Device;
 use ir::{self, BasicBlock, BBId, Dimension, InstId, Instruction, Operator};
-use ir::{AccessPattern, Operand, Size, Type, dim, mem};
+use ir::{AccessPattern, Operand, Size, Type, dim, mem, SparseVec};
 use ir::mem::Block;
 use itertools::Itertools;
 use std;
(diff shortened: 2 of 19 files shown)


src/ir/operand.rs
```diff
@@ -52,36 +110,39 @@ impl<'a> Operand<'a> {
...
+    pub fn new_float(val: Ratio<BigInt>, len: u16) -> Self {
         Float(val, len)
     }
 
     /// Renames a basic block id.
     pub fn merge_dims(&mut self, lhs: ir::DimId, rhs: ir::DimId) {
         match *self {
-            Inst(_, _, ref mut dim_map, _) |
+            Inst(_, _, ref mut dim_map, _) => { dim_map.merge_dims(lhs, rhs); },
             Reduce(_, _, ref mut dim_map, _) => { dim_map.merge_dims(lhs, rhs); },
```
ulysseB: The special case for Inst is not needed.

src/ir/operand.rs
```diff
@@ -90,7 +151,9 @@ impl<'a> Operand<'a> {
     /// Indicates if a `DimMap` should be lowered if lhs and rhs are not mapped.
     pub fn should_lower_map(&self, lhs: ir::DimId, rhs: ir::DimId) -> bool {
         match *self {
-            Inst(_, _, ref dim_map, _) |
```
ulysseB: The special case for Inst is not needed.

src/ir/dim_map.rs
```diff
@@ -5,55 +5,87 @@ use utils::*;
...
         filter_list(&mut self.map, f)
     }
 
     /// Returns true if the `DimMap` is empty.
-    pub fn is_empty(&self) -> bool { self.map.is_empty() }
+    pub fn is_empty(&self) -> bool {
+        self.map.is_empty()
+    }
+
+    pub fn lowering_map(&self, cnt: &mut ir::Counter) -> ir::operand::LoweringMap {
```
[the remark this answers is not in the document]
ulysseB: Also, I do not understand why does this needs to be done in `DimMap` instead of `LoweringMap::new`.

src/model/level.rs
```diff
@@ -243,25 +243,34 @@ pub struct DimMap {
...
-                    .collect::<VecSet<_>>();
-                if dst_dims.is_empty() || src_dims.is_empty() {
-                    None
-                } else {
-                    Some(DimMap { src, dst, src_dims, dst_dims })
-                }
-            },
+            ir::Operand::Inst(src, _, ref dim_map, _) =>
+                list_dim_map(space, src, dst, dim_map),
+            ir::Operand::Reduce(src, _, ref dim_map, _) =>
```
ulysseB: The special case for Inst is not needed.

src/ir/instruction.rs
```diff
@@ -17,28 +23,31 @@ impl std::fmt::Display for InstId {
...
         Ok(Instruction { operator, id, iter_dims })
     }
 
-    /// Returns the list of operands of an `Instruction`.
-    pub fn operands(&self) -> Vec<&Operand<'a>> { self.operator.operands() }
+    /// Returns an iterator over the operands of this instruction.
+    pub fn operands(&self) -> Vec<&Operand<'a, L>> { self.operator.operands() }
+
+    /// Returns a mutable iterator over the operands of this instruction.
+    pub fn operands_mut(&mut self) -> Vec<&mut Operand<'a, L>> { self.operator.operands_mut() }
```
ulysseB: I don't think this method is used. I would be in favor to remove it since it allows changing the operands of an instruction, which may violate some invariants (e.g change the type of the value produce by an instruction).

src/ir/function.rs
```diff
@@ -97,6 +102,15 @@ impl<'a> Function<'a> {
             self.mem_insts.push(id);
             self.mem_blocks.register_use(mem_id, id);
         }
+        Ok(inst)
+    }
+
+    /// Adds an instruction to the function.
+    pub fn add_inst(&mut self, op: Operator<'a, L>, iter_dims: HashSet<ir::DimId>)
```
ulysseB: Shouldn't we limit this function to non-freezed function ?

src/ir/instruction.rs
ulysseB: I think this method is not used. I would be in favor of removing it sinc it allows violating invariants (like changing the type of the value produced by an instruction).

src/model/level.rs
ulysseB: The special case for `Operand::Inst` is not needed.

src/ir/function.rs
```diff
@@ -266,15 +270,95 @@ impl<'a> Function<'a> {
...
+        let ir::Counter { next_mem, next_inst, next_dim } = counter;
+        for _ in mem_blocks.num_internal_blocks()..next_mem {
+            mem_blocks.push_lazy_tmp();
+        }
+        for _ in insts.len()..next_inst {
+            insts.push_lazy();
+        }
+        for _ in dims.len()..next_dim {
+            dims.push_lazy();
+        }
```
ulysseB: Your are allocating the holes in the sparse vectors. If I understand correctly, holes will be iterated (and skiped) each time the `SparseVec` is iterated. Thus I see no difference between having the holes at the end of the vector and having them dispersed in the vector. Why not directly allocate them when the operand is inserted in the `Function` instead of allocating them at the end ?

src/ir/function.rs
```diff
@@ -266,15 +270,95 @@ impl<'a> Function<'a> {
...
+            BBId::Inst(id) => &self.insts[id],
+            BBId::Dim(id) => self.dim(id),
+        }
+    }
+
+    /// Lists all `BasicBlock`s.
+    pub fn blocks<'b>(&'b self) -> impl Iterator<Item=&'b BasicBlock<'a>> {
+        self.insts.iter().map(|x| x as _).chain(self.dims().map(|x| x as _))
+    }
+
```
ulysseB: Why are `fn block` and `fn blocks` only defined for freezed functions ?

src/ir/dim_map.rs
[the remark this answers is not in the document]
Elarnon: Yes, that is what it does. It is in `DimMap` because of how that code came to be, initially `DimMap` owned the lowering map. I will change it so that `LoweringMap::new` takes the `DimMap` as argument instead.

src/ir/function.rs
Elarnon: Yes, I thought I did this actually.

src/ir/function.rs
Elarnon: Hm, that is true. Actually I will change it so that we dynamically allocate the holes when `init_lazy` is called, which is what we discussed initially.

src/ir/function.rs
Elarnon: Because the `BasicBlock` trait is parameterized on the lowering information, and I only implement the `BasicBlock<'a, LoweringMap>` for `Dimension<'a>`. If I don't do this, then rust gets confused all the time `as_dim` is called because it doesn't know whether to call `BasicBlock<'a, ()>::as_dim` or `BasicBlock<'a, LoweringMap>::as_dim`, even though both are identical. It is only used for frozen functions anyways.

src/ir/instruction.rs
Elarnon: It was used at some point I think. I will remove it.

src/ir/operand.rs
Elarnon: It was when `DimMap` had the lowering info builtin. Will revert.

src/ir/operand.rs
Elarnon: Same as above

src/model/level.rs
Elarnon: Same as above

Elarnon:
@ulysseB can you do another pass on this? There has been some code changes in part due to the `static_dims` changes which I'd like you to take another look at. Thanks!

src/codegen/namer.rs
```diff
@@ -164,7 +164,8 @@ impl<'a, 'b> NameMap<'a, 'b> {
                 Cow::Owned(self.namer.borrow().name_int(val, len)),
             Operand::Float(ref val, len) =>
                 Cow::Owned(self.namer.borrow().name_float(val, len)),
-            Operand::Inst(id, _, ref dim_map, _) |
+            Operand::Inst(id, _, ref dim_map, _) =>
+                Cow::Borrowed(self.name_inst_id(id, dim_map)),
```
ulysseB: You missed a `Inst`/`Reduce` code duplication to remove.

Elarnon: dang

Elarnon:
bors r=ulysseB

PR #98, opened 2018-08-01
Error Kind pattern for Lexer Interface

adjivas:
TelamonGen errors interface can be simplified.
This PR is a reorganization of TelamonGen/{Error,ast/Error} with failure trait.
It's replace the standard std::error::Error trait.


telamon-gen/src/lexer/mod.rs
```diff
@@ -49,50 +49,48 @@ pub type LexerItem = ParserSpanned<Token, Position, LexicalError>;
...
         }
     }
 }
 
+#[derive(Debug, Clone, PartialEq)]
+pub struct LexicalError {
+     pub cause: Spanned<ErrorKind>
+}
+
+impl Error for LexicalError {
```
ulysseB: the `failure` crate can derive this automatically for you. We already use it in other parts of Telamon so I would advise you tu use it here too.

telamon-gen/src/lexer/mod.rs
```diff
@@ -49,50 +49,48 @@ pub type LexerItem = ParserSpanned<Token, Position, LexicalError>;
...
+    fn description(&self) -> &str {
+        self.cause.data.description()
+    }
+}
+
```
ulysseB: You can also use failure here.

telamon-gen/src/error.rs
```diff
@@ -59,13 +59,22 @@ impl <'a>From<(path::Display<'a>,
                         Position { position: e, .. })
             } |
             ParseError::User {
-                error: LexicalError::UnexpectedToken(Position { position: l, .. }, .., Position { position: e, .. })
-            } |
-            ParseError::User {
-                error: LexicalError::InvalidInclude(Position { position: l, .. }, .., Position { position: e, .. })
+                error: LexicalError {
```
ulysseB: This match is becoming bloated with the position handling. Can you try to reduce the number of nested patterns ? Here are a few ideas:
- separate the span and the cause in LexicalError, you could even use directly `Span` in `LexicalError`.
- access to memberse using `mystruct.member` instead of matching them.

telamon-gen/src/ast/choice/enumeration.rs
```diff
@@ -159,7 +159,12 @@ impl EnumDef {
             if self.variables.len() != 2 {
                 Err(TypeError::BadSymmetricArg {
                         object_name: self.name.to_owned(),
-                        object_variables: self.variables.to_owned()
+                        object_variables: self.variables.iter().map(|variable| {
+                            let name: &String = variable.name.data.deref();
+                            let set_name: &String = variable.set.name.deref();
+                            (variable.name.with_data(name.to_owned()),
```
ulysseB: wouldn't `variable.name.clone()` be simpler ?

telamon-gen/src/ast/choice/enumeration.rs
```diff
@@ -176,7 +181,12 @@ impl EnumDef {
                     if name != rhs_name {
                         Err(TypeError::BadSymmetricArg {
                             object_name: self.name.to_owned(),
-                            object_variables: self.variables.to_owned()
+                            object_variables: self.variables.iter().map(|variable| {
+                                let name: &String = variable.name.data.deref();
+                                let set_name: &String = variable.set.name.deref();
+                                (variable.name.with_data(name.to_owned()),
```
ulysseB: same as above

telamon-gen/src/error.rs
ulysseB: The match would be simpler to understand if you took references to ben and end instead of l and e

telamon-gen/src/error.rs
```diff
@@ -84,21 +93,24 @@ impl <'a> fmt::Display for ProcessError<'a> {
...
-                    error: LexicalError::InvalidToken(_, ref token, _)
                 }), ..} => {
                 if let Some(span) = span {
                     write!(f, "{}, {} -> {}", token, span, path)
                 } else {
                     write!(f, "{} -> {}", token, path)
                 }
            },
+           ProcessError { path, span, cause: Cause::Parse(ParseError::User {
+                error: LexicalError{
```
ulysseB: the match would be simpler to understand if you did not unwrapped cause

telamon-gen/src/ast/choice/enumeration.rs
adjivas: Because _variable.name_ is already a string and I just need to borrow it than clone it.

telamon-gen/src/error.rs
adjivas: Solved

telamon-gen/src/lexer/mod.rs
adjivas: Solved

telamon-gen/src/lexer/mod.rs
adjivas: Solved

telamon-gen/src/ast/choice/enumeration.rs
ulysseB: Not sure I understand. Doesn't `variable.name` has the type `Spanned<String>` ? Arn't you cloning it with 'name.to_owned()' ?

adjivas: Yes, `variable.name` has the type `Spanned<String>`.
`to_owned` method usually clone but not always, that also can be a owned of the data.

telamon-gen/src/ast/choice/enumeration.rs
```diff
@@ -158,8 +158,13 @@ impl EnumDef {
         {
             if self.variables.len() != 2 {
                 Err(TypeError::BadSymmetricArg {
-                    object_name: self.name.to_owned(),
-                    object_variables: self.variables.to_owned(),
+                        object_name: self.name.to_owned(),
+                        object_variables: self.variables.iter().map(|variable| {
+                            let name: &String = variable.name.data.deref();
+                            let set_name: &String = variable.set.name.deref();
+                            (variable.name.with_data(name.to_owned()),
```
adjivas: rewrite

telamon-gen/src/ast/choice/enumeration.rs
```diff
@@ -185,7 +190,12 @@ impl EnumDef {
                     if name != rhs_name {
                         Err(TypeError::BadSymmetricArg {
                             object_name: self.name.to_owned(),
-                            object_variables: self.variables.to_owned(),
+                            object_variables: self.variables.iter().map(|variable| {
+                                let name: &String = variable.name.data.deref();
+                                let set_name: &String = variable.set.name.deref();
+                                (variable.name.with_data(name.to_owned()),
+                                 set_name.to_owned())
```
adjivas: rewrite

PR #100, opened 2018-08-02
Expose SearchSpace through the C API

Elarnon:
This add additional wrappers to the C API for creating a SearchSpace
from an IR function, as well as `telamon_explore` allowing to run the
search from the C API. There are still some limitations, notably:

 - Action objects can be constructed, but the constants for creating
   enum values (DimKind, etc.) are not exposed yet because cbindgen
   does not parse associated constants.

 - The exploration is currently "all or nothing", i.e. there is no way
   to exerce more precise control on what happens in the exploration or
   drive the exploration from the C side.

Those will be addressed in follow-up PRs.

### DIFF
--- a/src/ir/instruction.rs
+++ b/src/ir/instruction.rs
@@ -7,6 +7,7 @@ use utils::*;
 /// Uniquely identifies an instruction.
 #[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
 #[repr(C)]
+/// cbindgen:field-names=[id]
 pub struct InstId(pub u32);
 
 impl Into<usize> for InstId {
--- a/telamon-capi/src/explorer.rs
+++ b/telamon-capi/src/explorer.rs
@@ -1,0 +1,72 @@
+use std;
+use std::ffi::CStr;
+
+use libc::c_char;
+use telamon::explorer::{self, Config};
+
+use super::error::TelamonStatus;
+use super::search_space::SearchSpace;
+use super::Context;
+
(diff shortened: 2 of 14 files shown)


src/ir/instruction.rs
```diff
@@ -7,6 +7,7 @@ use utils::*;
 /// Uniquely identifies an instruction.
 #[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
 #[repr(C)]
+/// cbindgen:field-names=[id]
```
ulysseB: Is this kind of statement documented somewhere ?

telamon-capi/src/explorer.rs
```diff
@@ -0,0 +1,73 @@
...
+/// The resulting config object must be freed using
+/// `telamon_config_free`.
+#[no_mangle]
+pub extern "C" fn telamon_config_new() -> *mut Config {
+    Box::into_raw(Box::new(Config::default()))
+}
+
+/// Frees an explorer configuration.
+#[no_mangle]
+pub unsafe extern "C" fn telamon_config_free(config: *mut Config) -> TelamonStatus {
```
ulysseB: Why return a status ?

src/ir/instruction.rs
Elarnon: In cbindgen source code... I should make a PR to document them someday.

telamon-capi/src/explorer.rs
Elarnon: Because at some point I was failing if `config` was null. I'll remove it.

Elarnon:
bors r=ulysseB

PR #101, opened 2018-08-03
Fix C API types

Elarnon:
- Do not expose explorer configuration as `repr(C)` because it contains
   strings and options which can't be directly accessed in C. This is a
   temporary fix until we write (a set of macros to generate) proper C
   bindings for the configuration.

 - Since multiple enums can have constructors with identical names, we
   now prefix the constructor name with the enum name.


Elarnon:
bors r=ulysseB

PR #110, opened 2018-08-25
Include advanced test for Set 

adjivas:
@ulysseB,
You have asked to me a test the include keyword for three _Set_ named _A_, _B_ and _C_.
This names are lexed like _ValueIdent_ than _SetIdent_, that why the parser repported a unexpected token for the first Set. The error wasn't the include keyword but the name of set.

That why this PR add a checkcase to proof your test but with the variant _Aa_, _Bb_ and _Cc_.
I have also reorganized my _telamon-gen_ directory of test.

There are not more modifications implicated.
We can consider to relax the identification of a _SetIdent_.


adjivas:
bors test

adjivas:
bors try

ulysseB:
The error came from master. Should now be fixed.

bors r+

PR #112, opened 2018-08-31
Filters where not run when some sets where too small

ulysseB:
Telamon-gen generates filters to propagate the constraints. When the inputs of one filter are included in the inputs of another, they are merged together to improve propagation strength. However, if the inputs of the second depends on more sets of objects, a copy of the original filter must be kept so it is run when the second filter is not because some set is empty (or does not contain enough values). We currently do not keep the original filter.

This PR introduced a new test that exhibited this behavior and fixes it by checking the filters depend on the same sets before setting the flag that discards the first set. The inclusion between the lists of sets is already tested so we only need to make sure they have the same size.

### DIFF
--- a/telamon-gen/src/flat_filter.rs
+++ b/telamon-gen/src/flat_filter.rs
@@ -69,6 +69,7 @@ impl FlatFilter {
 
     /// Try to merge another filter into `Self`.
     fn try_merge(&mut self, other: &FlatFilter, ir_desc: &ir::IrDesc) -> bool {
+        // FIXME: do not set the merged flag if the set of foralls is bigger
         if self.inputs.is_empty() != other.inputs.is_empty() {
             return false;
         }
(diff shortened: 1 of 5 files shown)


telamon-gen/src/flat_filter.rs
```diff
@@ -109,11 +109,16 @@ impl FlatFilter {
             // Adapt and merge the rules.
             self.rules
                 .extend(other.rules.iter().map(|r| r.adapt(&adaptator)));
-            // Early exit for static filters.
-            if self.inputs.is_empty() {
-                return true;
+            // If the number of variable is different, don't set the merged flag: otherwise, the
+            // filter might not be run if their is not enought to objects to run the filter with
```
Elarnon: enought -> enough

Also I think it makes sense to copy the "The inclusion between the lists of sets is already tested so we only need to make sure they have the same size." explanation here.

Why is the check for static filters only when the sets of variables are not identical?

ulysseB:
bors retry

PR #116, opened 2018-09-05
Ensure mapped dimensions have the same size

ulysseB:
This PR adds a list of dimensions mapped together to the function. Pairs of dimensions in `dim_map` fields in the operands must now be mapped. A new constraint ensure mapped dimensions have the same size.

Mapped dimensions should also help us improve the data flow encoding: it exposes the structure of `dim_map`s to the constraint programing framework.

### DIFF
--- a/src/ir/function.rs
+++ b/src/ir/function.rs
@@ -339,6 +339,11 @@ impl<'a, L> Function<'a, L> {
     pub fn dim_mapping(&self, id: ir::MappedDimsId) -> &ir::MappedDims {
         &self.mapped_dims[id]
     }
+
+    /// Tries to find a mapping between two dimensions.
+    pub fn find_mapping(&self, lhs: ir::DimId, rhs: ir::DimId) -> Option<ir::MappedDimsId> {
+        self.dim(lhs).dim_mappings().find(|&id| self.dim_mapping(id).dims().contains(&rhs))
+    }
 }
 
 impl<'a> Function<'a, ()> {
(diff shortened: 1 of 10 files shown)


src/ir/function.rs
```diff
@@ -498,60 +557,62 @@ impl<'a> Function<'a> {
...
         ));
         self.insts.set_lazy(lowered.load, ld);
         self.insts[dst_inst].lower_dim_map(dst_operand_pos, lowered.load, ld_dim_map);
 
         Ok(lowered)
     }
+
+    /// Activate dimensions mapped to existing ones. Takes as input a list of mappings, in the
+    /// format `(mapping id, [old dim, new dim])`. Returns the mapping from old dimensions to new
+    /// ones if `old_to_new` is true. The mapping from new dimensions to old ones otherwise.
```
nicoTolly: Elaborate on this comment : mappings are symmetric, so it's weird that we have to care about the order in which they are returned. This is important  for the functions that take this mapping as input

nicoTolly: Also, the name of the function is a bit weird. Maybe create_mapped_dims ? Or promote_mapped_dims ? Could we find some verb better than "activate" for "this potential dimension that was referenced is now promoted to a concrete dimension and exists in the model" ? Don't really have a great suggestion on that

ulysseB: I don't have a better solution either, so I will just explain in details what is does in the comment

PR #119, opened 2018-09-07
Rely on Logical Dimensions to Constrain the Size of Merged Dimensions

ulysseB:
Changes how we ensure merged dimensions have the same size. We used to call `is_compatible_with` method on dimension sizes, but this method won't work when we enable strip-mining: we can't test the equality on `PartialSize` as they will only be partially specified. Instead we ensure logical dimensions have the same size and that the tiling factor is the same.

We only make changes for dimensions with a dynamic size. Dimensions with a static size where already covered.

This PR also fixes a small bug in TelamonGen. Some code was inside a loop when it should not have been. The result was that it was not run when the loop had 0 iterations and generated invalid code in that case.

### DIFF
--- a/telamon-gen/src/ir/mod.rs
+++ b/telamon-gen/src/ir/mod.rs
@@ -260,10 +260,9 @@ impl IrDesc {
         (arg_foralls, set_constraints, adaptator)
     }
 
-    /// Generates the list of sets to iterate and to constraints to iterate on the given
-    /// environement, but from the point of view of the given choice instance. The new
-    /// foralls iterating on current arguments are returned in a separate list than the
-    /// ones issued from foralls.
+    /// Generates the foralls and the set constraints to iterate on the given environement,
+    /// from the point of view of the given choice. Returns the foralls issued from arguments
+    /// in a different list than the foralls issued from foralls in the original environement.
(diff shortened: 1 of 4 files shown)


telamon-gen/src/ir/mod.rs
```diff
@@ -260,10 +260,9 @@ impl IrDesc {
         (arg_foralls, set_constraints, adaptator)
     }
 
-    /// Generates the list of sets to iterate and to constraints to iterate on the given
-    /// environement, but from the point of view of the given choice instance. The new
-    /// foralls iterating on current arguments are returned in a separate list than the
-    /// ones issued from foralls.
+    /// Generates the foralls and the set constraints to iterate on the given environement,
```
nicoTolly: Environment (typo)

telamon-gen/src/ir/mod.rs
```diff
@@ -260,10 +260,9 @@ impl IrDesc {
...
+    /// from the point of view of the given choice. Returns the foralls issued from arguments
+    /// in a different list than the foralls issued from foralls in the original environement.
```
nicoTolly: Same

ulysseB:
bors retry

PR #120, opened 2018-09-08
Limit the Size of Inner Block Dimensions

### LINKED ISSUE #45: Block dimensions are not checked against CUDA maximal block dim size.
Cuda limits the size of block dimensions. For `x`, the maximal size is `2^31-1` but for `y` and `z` it is `2^16-1`. Thus w can easily hit the limit on `y` and `z` axis.

The problem is that we do not statically know the size of potential block dimensions. Thus, we will need to find a way to express the maximal size they can take.

ulysseB:
Fixes #45.

CUDA imposes a limit on the size of BLOCK dimensions on y and z axis. This PR enforces this limit. It also fixes a bug in Telmon-gen: we forgot to rename some variables when applying a renaming.

### DIFF
--- a/src/device/cuda/gpu.rs
+++ b/src/device/cuda/gpu.rs
@@ -421,6 +421,10 @@ impl device::Device for Gpu {
         3
     }
 
+    fn max_inner_block_size(&self) -> u32 {
+        65535
+    }
+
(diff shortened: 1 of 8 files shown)


ulysseB:
bors try

PR #126, opened 2018-09-11
Generate a mapping value/name in NameMap

nicoTolly:
We want to refer directly to value in code generation

### DIFF
--- a/src/codegen/namer.rs
+++ b/src/codegen/namer.rs
@@ -34,6 +34,8 @@ pub trait Namer {
 pub struct NameMap<'a, 'b> {
     /// Provides fresh names.
     namer: std::cell::RefCell<&'b mut Namer>,
+    /// Keeps track of the names of the values used in the kernel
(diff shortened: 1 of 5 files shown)


src/codegen/namer.rs
```diff
@@ -142,6 +151,12 @@ impl<'a, 'b> NameMap<'a, 'b> {
         name_map
     }
 
+    /// Returns a name for a value
+    fn name_value(val: &ir::Value) -> String {
+        let id: usize = val.id().into();
+        format!("val_{}", id)
+    }
+
```
ulysseB: This won't work because the values are not declared in the `Namer`, so they won't be declared in the generated code. Use `namer.name(<type>)` to generate the names.

For that you will need a lowered type. The best for that is to create a wrapper around `ir::Value` (see `codegen::function::Instruction` for an example) that exposes a lowered type.

ulysseB:
bors retry

nicoTolly:
bors retry

PR #128, opened 2018-09-17
Added a Value Operand in ir

nicoTolly:
Create a new Operand to represent a Value in instruction

### DIFF
--- a/src/ir/operand.rs
+++ b/src/ir/operand.rs
@@ -111,6 +111,8 @@ pub enum Operand<'a, L = LoweringMap> {
     Reduce(InstId, Type, DimMap, Vec<ir::DimId>),
     /// A variable increased by a fixed amount at every step of some loops.
     InductionVar(ir::IndVarId, Type),
+    /// A value that can be read from by an instruction
+    Value(ir::Value),
 }
 
 impl<'a, L> Operand<'a, L> {
@@ -122,6 +124,7 @@ impl<'a, L> Operand<'a, L> {
             Addr(mem) => ir::Type::PtrTo(mem.into()),
             Index(..) => Type::I(32),
             Param(p) => p.t,
+            Value(ref val) => val.t(),
             Inst(_, t, ..) | Reduce(_, t, ..) | InductionVar(_, t) => t,
         }
     }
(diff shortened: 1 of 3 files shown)


src/ir/operand.rs
```diff
@@ -111,6 +113,8 @@ pub enum Operand<'a, L = LoweringMap> {
     Reduce(InstId, Type, DimMap, Vec<ir::DimId>),
     /// A variable increased by a fixed amount at every step of some loops.
     InductionVar(ir::IndVarId, Type),
+    /// A value that can be read from by an instruction
+    Value(ir::Value),
```
ulysseB: Why use `ir::Value` instead of `ir::ValueId` ? Here you duplicate the value instead of referencing it with an Id. Something like `Value(ir::ValueId, ir::Type)` would be more appropriate.

Also add a constructor `fn new_value(&ir::Value) -> ir::Operand` to the operand type so that the type field is automatically set to the right value (see new_inst).

src/ir/value.rs
```diff
@@ -63,6 +73,17 @@ impl ValueDef {
         function.inst_mut(*inst_id).set_result_value(self_id);
     }
 }
+
+#[derive(Clone, Debug, Hash, PartialEq, Eq)]
```
ulysseB: Why do you need to define a new struct instead of just using `ir::InstId` ?

src/ir/instruction.rs
```diff
@@ -56,6 +56,15 @@ impl<'a, L> Instruction<'a, L> {
         self.operator.operands()
     }
 
+    pub fn update_value_usepoints(&mut self) {
```
ulysseB: Here, you update the usepoints within the operand, but not in the value in `ir::Function`.

ulysseB:
Also, add a constraint to order def-points before use-points

src/ir/function.rs
```diff
@@ -399,6 +406,13 @@ impl<'a, L> Function<'a, L> {
         }
         mapping
     }
+    /// Returns true if inst2 depends on inst1 (check based on value)
+    pub fn is_dependency_of(&self, inst1_id: InstId, inst2_id: InstId) ->  bool {
+        match self.insts[inst1_id].result_value() {
+            None => true,
```
ulysseB: Don't you meant `None => false` ?

ulysseB:
By the way, the `create_inst_value` function in the builder should not create a new value if the instruction already has one. It should return the old one instead.

ulysseB:
bors retry

PR #135, opened 2018-09-24
Values handled in model, added dim_map to value operand

nicoTolly:
Value Operand in model. Had to had dim map to Operand::Value (not sure about that)


ulysseB:
bors try

nicoTolly:
bors retry

PR #139, opened 2018-09-30
Refactor values

ulysseB:
This PR refactors the IR code around values:
- it exposes uses and defs in sets
- makes cosmetic changes to value creation

### DIFF
--- a/src/ir/dimension.rs
+++ b/src/ir/dimension.rs
@@ -126,6 +126,10 @@ impl<'a> Dimension<'a> {
     }
 }
 
+lazy_static! {
+    static ref NO_VALUES: VecSet<ir::ValueId> = VecSet::default();
+}
+
 impl<'a> Statement<'a> for Dimension<'a> {
     fn stmt_id(&self) -> ir::StmtId {
         self.id.into()
@@ -133,6 +137,14 @@ impl<'a> Statement<'a> for Dimension<'a> {
 
     fn as_dim(&self) -> Option<&Dimension<'a>> {
         Some(self)
+    }
+
+    fn def_values(&self) -> &VecSet<ir::ValueId> {
+        &NO_VALUES
+    }
+
+    fn used_values(&self) -> &VecSet<ir::ValueId> {
+        &NO_VALUES
     }
 }
 
--- a/src/ir/dimension.rs
+++ b/src/ir/dimension.rs
@@ -127,6 +127,9 @@ impl<'a> Dimension<'a> {
 }
 
 lazy_static! {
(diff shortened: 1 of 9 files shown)


ulysseB:
Note to self: make new_objs works with new defs and use points

src/ir/dimension.rs
```diff
@@ -126,6 +126,10 @@ impl<'a> Dimension<'a> {
     }
 }
 
+lazy_static! {
```
nicoTolly: I don't know why you are using a static ref here and it looks like it is intended for future uses : anyway, unless it is completetly temporary I think this requires a comment explaining why it is here

nicoTolly: Ok, I think I do understand. This ref makes sure that we always return an empty set for def and used_values of a dimension (as these notions are irrelevant here). We could have created an empty VecSet at each call but this would have required useless allocation and initialization. I still think this should be stated in a comment

src/ir/statement.rs
```diff
@@ -30,12 +31,20 @@ impl From<ir::DimId> for StmtId {
...
     fn as_inst(&self) -> Option<&ir::Instruction<'a, L>> {
         None
     }
+
     /// Returns 'self' if it is a dimension
     fn as_dim(&self) -> Option<&ir::Dimension<'a>> {
         None
     }
+
+    /// Lists the values defined at this statement.
```
nicoTolly: There could be several values defined in the same statement ?

src/search_space/values.exh
```diff
@@ -8,6 +8,36 @@ set Values:
   new_objs = "$objs.values"
 end
 
+set DefStatements($value in Values) subsetof Statements:
+  item_type = "ir::Statement"
+  id_type = "ir::StmtId"
+  item_getter = "$fun.statement($id)"
+  id_getter = "$item.id()"
+  iterator = "$value.def_points().map(|id| $fun.statment(id)"
```
nicoTolly: unclosed parenthese + should be statement instead of statment

src/ir/dimension.rs
ulysseB: Done.

src/ir/statement.rs
ulysseB: Currently no, but once we have dim maps yes: if v0 = DimMap(v1), then v0 and v1 have the same def.

src/search_space/values.exh
ulysseB: good catch. Thanks.

ulysseB:
bors retry

PR #141, opened 2018-10-03
Reorder choices

nicoTolly:
Mini PR adding a function list_with_ordering. We should now make possible to specify an order choice in config file

### DIFF
--- a/src/explorer/choice.rs
+++ b/src/explorer/choice.rs
@@ -8,6 +8,18 @@ use search_space::{Action, Domain, NumSet, Order, SearchSpace};
 // TODO(search_space): explore and lower loayouts directly from the regular actions.
 pub type Choice = Vec<ActionEx>;
 
+/// A list of ChoiceGroup representing the order in which we want to determine choices
+pub struct ChoiceOrdering(Vec<ChoiceGroup>);
+
+/// An enum listing the Group of choices we can make
+/// For example, we can make first all DimKind decisions, then all Order decisions, etc.
+pub enum ChoiceGroup {
+    DimKind,
+    DimMap,
+    Order,
+    MemSpace,
+    InstFlag
+}
 /// Either a regular action or a manually applied action.
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub enum ActionEx {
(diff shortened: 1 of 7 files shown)


src/explorer/choice.rs
```diff
@@ -19,6 +31,70 @@ pub enum ActionEx {
     },
 }
 
+/// Given the order in which we want to explore the space, gives the list of choices still to be
+/// considered
+/// TODO: make this lazy
+pub fn list_with_ordering<'a>(
```
Elarnon: What we need is basically a generator, but that doesn't exist in Rust (yet).  I think the simplest solution to achieve this would be to define a `struct NestedChoiceIterator<I>(I)`, implement `impl<I: Iterator<Item = Box<dyn Iterator<Item = Choice>>>> Iterator for NestedChoiceIterator<I>` and return it as the `impl Iterator<Item = Choice>`:

```rust
NestedChoiceIterator::new(choice_order.0.iter().map(|group| -> Box<dyn Iterator<Item = Choice>> {
    Box::new(match group { .. })
}))
```

There may even be a chance that Rust is able to eliminate the extra allocations when we simply do a `.next()`.

src/explorer/choice.rs
```diff
@@ -19,48 +87,66 @@ pub enum ActionEx {
...
+    }
+}
+
+impl<I: Iterator> Iterator for NestedIterator<I>
+where
+    I::Item: Iterator,
+{
+    type Item = <I::Item as Iterator>::Item;
+
+    fn next(&mut self) -> Option<Self::Item> {
```
Elarnon: This may be premature optimisation, but we should use loops instead of recursion here as it should make it easier for rustc to simplify the common case where we have a single `next()` call.  Something like:

```rust
fn next(&mut self) -> Option<Self::Item> {
    loop {
        if let Some(ref mut current_it) = self.current_local_iterator {
            if let Some(choice) = current_it.next() {
                break Some(choice);
            }
        }

        // If we are here ...
        if let Some(local_it) = self.glob_iterator.next() {
            self.current_local_iterator = Some(local_it)
        } else {
            break None
        }
    }
}
```
}

src/explorer/choice.rs
```diff
@@ -19,48 +87,66 @@ pub enum ActionEx {
...
-                })
-        }))
-        .chain(fun.mem_insts().flat_map(move |inst| {
-            let flags = space.domain().get_inst_flag(inst.id()).list();
-            gen_choice(flags, &|f| Action::InstFlag(inst.id(), f))
-        }))
+    DEFAULT_ORDERING.list(space)
+}
+
+struct NestedIterator<I: Iterator>
```
Elarnon: Please add a couple lines to document this type and either move it to top of the file (close to where it actually gets used). Or even as a local type inside the `ChoiceOrdering::list` function maybe?

PR #142, opened 2018-10-05
Implement Statement for non-forzen instructions and dimensions (+random fixes)

ulysseB:
This PR implements Statement for not yet frozen instructions and dimensions. This allows us to manipulate then in a uniform way when needed. It also rename statement accessors from `block` to `statement` and make public the return type of binary operators.

### DIFF
--- a/src/ir/dimension.rs
+++ b/src/ir/dimension.rs
@@ -1,6 +1,6 @@
 //! Represents iteration dimensions.
 use ir::{self, Statement};
-use std::fmt;
+use std;
 use utils::*;
 
 /// Provides a unique identifier for iteration dimensions.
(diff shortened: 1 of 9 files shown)


src/ir/dimension.rs
```diff
@@ -17,31 +17,48 @@ impl Into<usize> for DimId {
...
+impl std::fmt::Display for DimId {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
         self.0.fmt(f)
     }
 }
 
 /// Represents an iteration dimension.
 #[derive(Clone, Debug)]
-pub struct Dimension<'a> {
+pub struct Dimension<'a, L = ir::LoweringMap> {
```
nicoTolly: Need to comment sowhere what these L/LoweringMap/() are about (differenciating lowered from non-lowered function)

ulysseB: Done, in ir::Function.

PR #148, opened 2018-10-12
Fixed wrong zero in local_selection

nicoTolly:
Problem was that api::pick_index returns 0 unconditionally when it should actually return the index of the first non-tested child (which is the first child in the list passed to pick_index, but not the first in the node of the tree)


nicoTolly:
bors r=elarnon

PR #149, opened 2018-10-15
Improve vectorization decisions expressiveness.

ulysseB:
This commit adds two features:
- It exposes two levels of vectorization instead of one (INNER_VECTOR
  and OUTER_VECTOR dimension kinds).
- It allows multiple dimensions to be mapped to the same level. This
  is usefull when we have multiple small dimensions and wants to map
  them to the same big hardware dimension.

It also fixes a test that was using dimensions with no instructions
inside, which is not supported by Telamon.


nicoTolly:
bors retry

PR #151, opened 2018-10-17
Add `Last` and `DimMap` variables.

ulysseB:
This PR enhance variables to
- take the last value produced in a loop nest and
- handle point-to-point communication
This was previously directly handled by instructions operands.

This changes code in sevral places
- in ir/ to support the new kinds of variables
- in codegen/ to correctly assign names to variables and alias them when
  a variable is defined as the value of another.
- in model/ to compute data dependencies
- in search_space/ to enfore constraints on mapped dimensions. Most of
  the changes is moving code from choices.exh to variable.exh


src/ir/variable.rs
```diff
@@ -21,16 +20,43 @@ pub struct Variable {
...
+/// This is usefull to limit the size of the search space by removing useless decisions.
+/// For example, we don't want to store in memory the operand of a store.
+#[derive(Clone, Copy, Debug, PartialOrd, Ord, PartialEq, Eq)]
+pub enum MemoryLevel {
+    /// The variable must be stored in registers and the producer and consumer must not be
+    /// separated by synchronisations.
+    RegisterNoSync,
+    /// The variable must be stored in registers.
+    Register,
+    /// The variable must be stored in registers or a local, fast, memroy.
```
nicoTolly: Typo

src/ir/variable.rs
```diff
@@ -21,16 +20,43 @@ pub struct Variable {
...
+    memory_level: MemoryLevel,
+    dimensions: VecSet<ir::DimId>,
+    def_points: VecSet<ir::StmtId>,
     use_points: VecSet<ir::StmtId>,
 }
 
+/// Indicates the slowest memory level where a variable may be stored.
+///
+/// This is usefull to limit the size of the search space by removing useless decisions.
+/// For example, we don't want to store in memory the operand of a store.
```
nicoTolly: Maybe add an example of a variable that can be stored in fastMem but not in SlowMem ? I cannot think of any on the fly

src/ir/variable.rs
```diff
@@ -21,16 +20,43 @@ pub struct Variable {
...
+    pub fn new<L>(id: VarId, def: VarDef, fun: &ir::Function<L>) -> Self {
+        let t = def.t(fun);
+        let def_points = def.def_points(fun);
+        let dimensions = def.dimensions(fun);
         Variable {
             id,
             t,
             def,
+            // TODO(ulysse): allow lowering to memory.
+            memory_level: MemoryLevel::Register,
```
nicoTolly: What ? Is there a way to create a variable that can be store elsewhere than in Register ?

src/ir/variable.rs
```diff
@@ -52,70 +78,152 @@ impl Variable {
...
-    }
-
     /// Returns the type of the variable if used on the context of `function`.
-    pub fn t<L>(self, fun: &ir::Function<L>) -> ir::Type {
-        let VarDef::Inst(inst_id) = self;
-        unwrap!(fun.inst(inst_id).t())
+    pub fn t<L>(&self, fun: &ir::Function<L>) -> ir::Type {
+        match self {
+            VarDef::Inst(inst_id) => unwrap!(fun.inst(*inst_id).t()),
+            VarDef::DimMap(value_id, ..) | VarDef::Last(value_id, ..) => {
```
nicoTolly: value_id should be called var_id

src/ir/variable.rs
```diff
@@ -52,70 +78,152 @@ impl Variable {
...
+                fun.variable(*value_id).t()
```
nicoTolly: Don't understand why this is not a self-reference. Do we have the guarantee that this variable if defined as inst() ? Or am I missing something ?

src/codegen/variable.rs
```diff
@@ -0,0 +1,236 @@
...
+use ir;
+use search_space::*;
+use utils::*;
+
+/// Wraps an `ir::Variable` to expose specified decisions.
+pub struct Variable<'a> {
+    variable: &'a ir::Variable,
+    t: ir::Type,
+    instantiation_dims: HashMap<ir::DimId, usize>,
+    alias: Option<Alias>,
```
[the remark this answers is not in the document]
ulysseB: You can chain aliases: A alias with B alias with C. I added a comment to explain that on the public method.

src/codegen/variable.rs
```diff
@@ -0,0 +1,236 @@
...
+            other_variable,
+            dim_mapping,
+            reverse_mapping,
+        }
+    }
+}
+
+/// Add a wrapper around variables to expose specified decisions. Orders variables with an
+/// alias after the variable they depend on.
+pub fn wrap_variables<'a>(space: &'a SearchSpace) -> Vec<Variable<'a>> {
```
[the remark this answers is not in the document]
ulysseB: Done.

src/codegen/name_map.rs
```diff
@@ -106,7 +101,7 @@ impl<'a, 'b> NameMap<'a, 'b> {
                 mem_blocks.insert(mem_block.id(), name);
             }
         }
-
+        let variables = VariableNames::create(function, namer);
```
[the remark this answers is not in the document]
ulysseB: I aggree. However, if you don't mind, I will do that in a later PR. I expect a good part of the code to be removed with the old operand system (Inst, Mem and Reduce should all be replaced by Variable).

src/ir/variable.rs
ulysseB: A variable can't depend on itself.

src/ir/variable.rs
ulysseB: Currently no. My goal is to unify memory blocks and variables.

src/ir/variable.rs
ulysseB: Done.

PR #152, opened 2018-10-18
Enforce that all code must typecheck with Travis.

Elarnon:
There are some rarely used code paths (some benches and the C API) which
frequently gets broken by changes in Telamon APIs.  This patch fixes
them (for the time being) and adds a Travis check that all the tests,
benches, and binaries from all the crates in the repo must typecheck.


Elarnon:
bors try

PR #155, opened 2018-10-19
Add support for vector registers.

ulysseB:
This PR also makes a small fixes:
- in telamon-gen: universal quantifiers where not processed in the right order when
some quantified sets had parameters.
- in variables.exh: some set definitions where wrong but not used before


ulysseB:
bors retry

PR #157, opened 2018-10-22
Remove the memory ID for external memory blocks.

ulysseB:
This commit removes external arrays from the objects exposed to the
constraint satisfaction problem. This ends up simplifying or removing
useless code, which will simplify later changes.


src/ir/mem.rs
```diff
@@ -5,67 +5,31 @@ use utils::*;
...
-}
-
 #[derive(Clone, Copy, PartialEq, Eq, Hash, Debug, Serialize, Deserialize)]
 #[repr(C)]
 /// cbindgen:field-names=[id]
-pub struct InternalId(pub u32);
+pub struct MemId(pub u32);
 
-impl Into<usize> for InternalId {
+impl Into<usize> for MemId {
```
nicoTolly: According to the rust doc, implementing From should be preferred to implementing Into when possible. So, replace this by impl From<MemId> for usize

src/ir/mod.rs
```diff
@@ -295,6 +293,15 @@ where
     }
 }
 
+impl<I, T> Default for SparseVec<I, T>
+where
```
nicoTolly: This where seems useless in the context

ulysseB: It is required by SparseVec.

ulysseB:
bors retry

ulysseB:
bors retry

ulysseB:
bors retry

PR #161, opened 2018-10-27
Fix Cuda Characterization

ulysseB:
This PR fixes two bugs:
- it handles the case where the `cuda_gpus.json` file was missing
- it emits a warning instead of failing when  the syncthread latency is not 0. This does not impact correctness but we could improve the model by taking this lantency into account.


ulysseB:
bors r_

PR #162, opened 2018-10-29
Expose the length and type of memory blocks.

ulysseB:
This PR exposes the length and element type of memory block in code generation instead of just providing a number of bytes to allocate.

Suggested review order:
- `src/ir/mem.rs`
- `src/codegen/function.rs`
- all the rest.

**Do not merge into master, merge in the variable branch instead**

### DIFF
--- a/src/codegen/function.rs
+++ b/src/codegen/function.rs
@@ -342,7 +342,7 @@ impl<'a> MemoryRegion<'a> {
         out
     }
 
-    /// Retruns the size of the memory block in number of elements accessible by each
+    /// Returns the size of the memory block in number of elements accessible by each
     /// thread.
     pub fn local_len(&self) -> &codegen::Size<'a> {
         &self.len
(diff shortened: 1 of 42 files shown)


src/ir/mem.rs
```diff
@@ -59,9 +59,14 @@ impl Block {
         true
     }
 
-    /// Return the base size of the block, if it is statically known.
-    pub fn base_size(&self) -> u32 {
-        self.base_size * unwrap!(self.elements_type.len_byte())
+    /// Return the base size of the block in bytes.
```
nicoTolly: What do you mean by "base size" ? Is it just the number of bytes in the block or is there any other subtlety ? If it is simply the number of bytes, remove the word base

src/codegen/function.rs
```diff
@@ -331,25 +331,41 @@ impl<'a> MemoryRegion<'a> {
...
         if let Some(ref s) = self.num_private_copies {
             out *= s
         }
         out
     }
 
-    /// Returns the size of the part of the allocated memory accessible by each thread.
-    pub fn local_size(&self) -> &codegen::Size<'a> {
-        &self.size
+    /// Retruns the size of the memory block in number of elements accessible by each
```
nicoTolly: Typo :)

src/ir/mem.rs
ulysseB: Added a comment to clarify.

ulysseB:
bors retry

PR #163, opened 2018-10-29
Add support for loop-carried variables.

ulysseB:
This PR adds support for loop-carried variables. They take the form of a `VarDef::Fby` operator.

Suggested review order:
- `src/ir/variables.rs`
- `src/search_space/*`
- `src/ir/*`
- all the rest.

**Merge into variable instead of master**

### DIFF
--- a/src/ir/statement.rs
+++ b/src/ir/statement.rs
@@ -26,6 +26,45 @@ impl From<ir::DimId> for StmtId {
     }
 }
 
+/// Either a `StmtId` or a mutable reference to a statement.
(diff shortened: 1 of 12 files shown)


src/ir/statement.rs
```diff
@@ -26,6 +26,45 @@ impl From<ir::DimId> for StmtId {
     }
 }
 
+/// Either a `StmtId` or a mutable reference to a statement.
```
nicoTolly: Can you comment here why we need such a thing ? It does not seem obvious

ulysseB:
bors retry

PR #164, opened 2018-10-31
Make fake context compatible with any underlying device

Elarnon:
This patch extract the FakeContext (and associated helper classes) from
the tests into telamon::device directly, and adds proper support for
dimension sizes (parameters) which was previously ignored due to the
test use case.  This makes it a general context for use when we want to
be able to manipulate candidates, but don't actually need to evaluate
the resulting implementations.

It can be used as follows:

```rust
extern crate serde_json;

use telamon::device::{cuda::Gpu, fake::FakeContext};

// cuda_gpus_path should contain the path to the cuda_gpus.json file to
// use.
let gpu: Gpu = unwrap!(serde_json::from_reader(
    &unwrap!(std::fs::File::open(cuda_gpus_path))));
let mut context = FakeContext::new(gpu);
```

telamon:
 * src/device/fake.rs: Added.
 * src/device/mod.rs: Expose the new `fake` module.
 * tests/common/fake.rs: Use the `FakeContext` from telamon::device.

### DIFF
--- a/src/device/fake.rs
+++ b/src/device/fake.rs
@@ -1,0 +1,107 @@
+use std::collections::HashMap;
+use std::marker::PhantomData;
+use std::sync::Arc;
+
+use codegen;
+use explorer::Candidate;
+use ir;
+
+use super::{
+    ArgMap, ArrayArgument, AsyncCallback, AsyncEvaluator, Context, Device, EvalMode,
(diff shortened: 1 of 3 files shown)


Elarnon:
Oops. bors try

Elarnon:
bors try

PR #166, opened 2018-10-31
Update the builder to use variables.

ulysseB:
This PR uses the new variable system everywhere where it is possible. It
also removes the old `Reduce` operand as it is not needed anymore and
fixes minor bugs caught in the process.

Suggested review order:
- src/ir/operand.rs
- src/helper/operand.rs
- src/helper/builder.rs
- src/*
- all the rest

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ branches:
     - master
 language: rust
 cache: cargo
-sudo: false
+sudo: required
 rust:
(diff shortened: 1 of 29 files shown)


ulysseB:
bors retry

ulysseB:
bors retry

ulysseB:
bors retry

PR #169, opened 2018-11-07
 Support CUDA_HOME environment variable 

Elarnon:
This patch makes it so that Telamon's build.rs uses the CUDA headers
from the user's CUDA_HOME environment variable if defined.  This matches
the behavior of most CUDA-based toolkits, including nVidia's own
examples.

The CUDA_HOME environment variable should point to the directory where
the CUDA toolkit is installed, which is /usr/local/cuda by default when
running nVidia's installer.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -21,13 +21,10 @@ fn compile_link_cuda() {
     let mut builder = cc::Build::new();
 
     // If CUDA_HOME is defined, use the cuda headers from there.
-    match env::var_os("CUDA_HOME").map(PathBuf::from) {
-        Some(cuda_home) => {
-            builder
-                .include(cuda_home.join("include"))
-                .include(cuda_home.join("extras").join("CUPTI").join("include"));
-        }
-        None => (),
+    if let Some(cuda_home) = env::var_os("CUDA_HOME").map(PathBuf::from) {
+        builder
+            .include(cuda_home.join("include"))
+            .include(cuda_home.join("extras").join("CUPTI").join("include"));
     }
 
     builder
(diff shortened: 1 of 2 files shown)


Elarnon:
bors try

build.rs
```diff
@@ -17,7 +18,19 @@ fn add_dependency(dep: &Path) {
 
 /// Compiles and links the cuda wrapper and libraries.
 fn compile_link_cuda() {
-    cc::Build::new()
+    let mut builder = cc::Build::new();
+
+    // If CUDA_HOME is defined, use the cuda headers from there.
+    match env::var_os("CUDA_HOME").map(PathBuf::from) {
```
ulysseB: When you only match on a single case, `if let Some(cuda_home) = ...` makes things more readable.

src/explorer/parallel_list.rs
```diff
@@ -58,6 +58,7 @@ impl<'a> ParallelCandidateList<'a> {
     }
 
     /// Insert a candidate to process.
+    #[allow(dead_code)]
     pub fn insert(&self, candidate: Candidate<'a>) {
```
ulysseB: Do we even need this method ?

Elarnon: I don't know but I am tired of the compilation warnings :)  I can remove it if you prefer.

build.rs
Elarnon: I was going to answer that I prefer `let builder = match ..` rather than `let builder = if let ..` but turns out I forgot that I refactored this.  Will change.

src/explorer/parallel_list.rs
ulysseB: Yes, just remove it,

Elarnon:
bors r=ulysseB

PR #173, opened 2018-11-21
Add object-safe trait for building candidates

Elarnon:
This patch introduces a `KernelBuilder` structure which wraps a kernel's
parameters.  The `KernelBuilder` provides a `scoped_kernel` helper
function which takes care of building the kernel itself and its
signature, and passes them on to a scoped closure[1].  Moreover, it
introduces a `ScopedCandidates` trait (and implements it for the
`KernelBuilder`) that allows building candidates from a type which
doesn't depend on the actual kernel used:

```rust
let builder: Box<dyn ScopedCandidates<'_, _, _, _>> = match KernelType {
	Axpy(params) => Box::new(linalg::Axpy<'_, f32>::builder(params)),
	MatMul(params) => Box::new(
		linalg::MatMul<'_, f32>::builder(params),
	)
}
```

This makes it easier to write kernel-independent tools around the core
Telamon library; in particular, this will allow more direct integration with the `Candidates` on the Python side of things (we want to reach the candidates from Python).

I am not 100% sure this is the right approach; ideally I'd like a way to create the Context (from Python), then interact with it by creating various candidates, contexts, etc. but as far as I can tell this is going to be a pain due to all the lifetime-bound references lying around.  Using a callback in that case seems like the less painful path to me; unless @ulysseB has a better suggestion.

1: The closure is required to ensure lifetime invariants relative to the
   mutable reference on the execution context.

telamon-kernels/
 * src/kernel.rs:
	Introduce the `KernelBuilder` and `ScopedCandidates`, and rewrite
	the `Kernel` helper functions (`benchmark`, etc.) to use them.
	Also introduce and use a more explicit `MemInit` enum for specifying
	memory initialization procedure instead of relying on a
	`random_fill` boolean.
 * src/lib.rs:
 * benches/cuda_bound.rs:
 * benches/cuda_search.rs:
 * benches/cuda_variance.rs:

telamon-capi/
 * telamon-capi/src/lib.rs

### DIFF
--- a/kernels/benches/cuda_bound.rs
+++ b/kernels/benches/cuda_bound.rs
@@ -4,7 +4,7 @@ extern crate telamon;
 extern crate telamon_kernels;
 
 use telamon::device::cuda;
-use telamon_kernels::{analyze_bounds, linalg, Kernel};
+use telamon_kernels::{analyze_bounds, linalg, Kernel, MemInit};
 
 fn main() {
     env_logger::init();
@@ -29,7 +29,7 @@ where
     K: Kernel<'a>,
 {
     let mut context = cuda::Context::new(executor);
-    let bounds = K::test_bound(params, num_runs, true, &mut context);
+    let bounds = K::test_bound(params, num_runs, MemInit::RandomFill, &mut context);
     println!("bounds for kernel {}", K::name());
     analyze_bounds(bounds);
 }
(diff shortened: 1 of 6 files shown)


Elarnon:
Actually I realized on the trip home that `device::Context` objects are not meant to be reused.  And we also only need immutable references to the executor.  In that case I think I should be able to tweak the lifetimes to get a Context + Candidates wrapper out and use that in the Python world.  I'll do a take 2 tomorrow.

PR #174, opened 2018-11-26
Make ArgMap object-safe

Elarnon:
This patch makes it so that `ArgMap` can be implemented for
`Box<dyn ArgMap>`, easing the compatibility with the C API.  This is
achieved by requiring devices to implement type-erased versions of
`bind_scalar` and `bind_array` (all existing implementations of `ArgMap`
were already type-erasing their arguments).

The pre-existing typed APIs are implemented as wrappers on their
type-erased version, allowing users (such as the `kernels` library) to
work with minor lifetime-related modifications.

telamon/
 * src/device/argument.rs:
       The `ScalarArgument` trait is revamped to:

        - Remove the `Copy` and `PartialEq` bounds, which were not used
          and prevented object safety by using `Self` in their methods

        - Guard the non-object-safe `t`, `gen_random` and `as_operand`
          methods with `Self: Sized` since we only use them when we know
          the exact type

        - Add `get_type` as an object-safe version of `t`

       All the old methods are still provided as before.

       In addition, the `write_array<S>` and `read_array<S>` functions
       are replaced with `write` and `read` methods on a new
       `ArrayArgumentExt` extension trait.

 * src/device/context.rs:
       The `ArgMap` trait is revamped:

        - `ArgMap` now takes a lifetime parameter bounding the lifetime
          of the generated arrays.  Most previous bounds were of the
          form `ArgMap + 'a`; they are now `ArgMap<'a>` instead.

        - The `bind_scalar` method is replaced with `bind_erased_scalar`
          which takes as argument a `ScalarArgument` trait object

        - The `bind_array` method is replaced with a `bind_erased_array`
          which takes as argument an `ir::Type` and returns an
          `ArrayArgument` trait object instead of the `Array` type
          parameter, which is removed.

       All the typed functionality of `ArgMap` (`bind_scalar` and
       `bind_array`) are automatically implemented through an
       `ArgMapExt` extension trait.

 * src/device/cuda/context.rs: Upgrade `ArgMap` implementation.
 * src/device/x86/context.rs: Upgrade `ArgMap` implementation.
 * src/device/fake.rs: Upgrade `ArgMap` implementation.
 * src/device/x86/cpu_argument.rs:
    Use `ScalarArgument` instead of the equivalent ` CpuScalarArg`
 * src/device/cuda/api/module.rs:
 * src/device/cuda/api/fake.rs:
 * src/device/cuda/characterize/gen.rs:
 * src/device/mod.rs:
 * src/helper/signature.rs:
 * src/helper/tensor.rs:
 * tests/cuda.rs:
 * tools/bench_perf_model/latency.rs:
 * tools/bench_perf_model/main.rs:
 * tools/bench_perf_model/memory.rs:
 * tools/bench_perf_model/tests.rs:

telamon-kernels/
 * benches/cuda_variance.rs:
 * src/kernel.rs:
 * src/lib.rs:
 * src/linalg.rs:

telamon-capi/
 * src/cuda.rs:
 * src/lib.rs:

### DIFF
--- a/src/device/argument.rs
+++ b/src/device/argument.rs
@@ -81,7 +81,6 @@ float_scalar_argument!(unsafe impl ScalarArgument for f64 [(0.) .. (1.)]);
 // `ScalarArgument` implementation for integer types.  Requires that the type is an integer scalar
 // type (see safety requirements for `ScalarArgument`).
 macro_rules! int_scalar_argument {
-
     (unsafe impl ScalarArgument for $ty:ident [ ($start:expr) .. ($stop:expr) ]) => {
         unsafe impl ScalarArgument for $ty {
             fn as_size(&self) -> Option<u32> {
--- a/src/device/argument.rs
+++ b/src/device/argument.rs
@@ -40,14 +40,17 @@ pub unsafe trait ScalarArgument:
         Self: Sized;
 }
 
+// Returns the size of a type in bits.  Used for the `ScalarArgument` implementations below.
 macro_rules! size_bits {
     ($ty:ty) => {
         8 * std::mem::size_of::<$ty>() as u16
     };
 }
 
-macro_rules! scalar_argument {
-    ($ty:ident float [ ($start:expr) .. ($stop:expr) ]) => {
+// `ScalarArgument` implementation for floating point types.  Requires that the type is a
+// floating-point scalar type (see safety requirements for `ScalarArgument`).
+macro_rules! float_scalar_argument {
+    (unsafe impl ScalarArgument for $ty:ident [ ($start:expr) .. ($stop:expr) ]) => {
         unsafe impl ScalarArgument for $ty {
             fn t() -> ir::Type {
                 ir::Type::F(size_bits!($ty))
@@ -70,8 +73,16 @@ macro_rules! scalar_argument {
             }
         }
     };
+}
 
-    ($ty:ident int [ ($start:expr) .. ($stop:expr) ]) => {
+float_scalar_argument!(unsafe impl ScalarArgument for f32 [(0.) .. (1.)]);
+float_scalar_argument!(unsafe impl ScalarArgument for f64 [(0.) .. (1.)]);
+
+// `ScalarArgument` implementation for integer types.  Requires that the type is an integer scalar
+// type (see safety requirements for `ScalarArgument`).
+macro_rules! int_scalar_argument {
+
+    (unsafe impl ScalarArgument for $ty:ident [ ($start:expr) .. ($stop:expr) ]) => {
         unsafe impl ScalarArgument for $ty {
             fn as_size(&self) -> Option<u32> {
                 if std::mem::size_of::<$ty>() <= std::mem::size_of::<u32>() {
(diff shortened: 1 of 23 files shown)


src/device/argument.rs
```diff
@@ -1,156 +1,115 @@
...
-    }
-
-    fn as_operand<L>(&self) -> ir::Operand<'static, L> {
-        ir::Operand::new_int(unwrap!(BigInt::from_i32(*self)), 32)
-    }
-
-    fn gen_random<R: Rng>(rng: &mut R) -> Self {
-        rng.gen_range(0, 100)
-    }
+macro_rules! scalar_argument {
```
ulysseB: You should specify under which condition the macro is correct (i.e. does not generate unsafe code). Also I am not a big fan of `unsafe impl` in a macro, but I don't have a better solution.

ulysseB: Maybe you could change the rules to be like:
```
(unsafe $ty:ident float [ ($start:expr) .. ($stop:expr) ]) => {
```

ulysseB: Plus verbeux, mais plus évocatif du code genéré:
```
(unsafe impl ScalarArgument for $ty:ident : float [ ($start:expr) .. ($stop:expr) ]) => ...
```

src/device/argument.rs
```diff
@@ -1,156 +1,115 @@
...
-    fn gen_random<R: Rng>(rng: &mut R) -> Self {
-        rng.gen_range(0, 100)
-    }
-}
+scalar_argument!(f32 float [(0.) .. (  1.)]);
+scalar_argument!(f64 float [(0.) .. (  1.)]);
+scalar_argument!(i8  int   [(0 ) .. ( 10)]);
+scalar_argument!(i16 int   [(0 ) .. (100)]);
+scalar_argument!(i32 int   [(0 ) .. (100)]);
+scalar_argument!(i64 int   [(0 ) .. (100)]);
```
ulysseB: Why are the parenthesis on the range bounds necessary ?

src/device/cuda/api/module.rs
```diff
@@ -160,3 +160,13 @@ where
         device::ScalarArgument::as_size(self)
     }
 }
+
+impl Argument for Box<dyn device::ScalarArgument> {
```
ulysseB: Why do you need this impl ?

Elarnon: Because `bind_erased_scalar` now gets a `Box<dyn device::ScalarArgument>` which we convert into an `Arc<dyn Argument>` to put it in the `parameters` hash map (same thing in `x86` module).  It should be possible to make `Argument` an `enum` instead to avoid the extra indirection, but I'd rather do it in a separate PR if needed.  I don't think this would impact performance enough to warrant it, however.

src/device/argument.rs
Elarnon: I don't like having `unsafe impl` in a macro that much either, but the previous copy-pasted code was far from optimal as well. 

My understanding was that the "macro syntax is evocative of the output" rule was meant for macros intended to have external users, not for internal "auto-impl" macros which are implementation details and generally very straightforward (see for example [`int_impl!` in `core`](https://doc.rust-lang.org/nightly/src/core/num/mod.rs.html#2053)).

I will update the macro to reference `unsafe` and add documentation that this implements `ScalarArgument` for the underlying type and thus has the same safety requirements.

Elarnon:
@ulysseB ready for a second round!

Elarnon:
Sigh, I guess I really need to set up that pre-commit `rustfmt` hook...

Elarnon:
bors retry

PR #175, opened 2018-11-29
Make the MCTS explorer more generic

Elarnon:
This patch tries to address various issues with the MCTS explorer, which
is outgrowing its initially planned feature set.  In particular, it
makes it more generic w.r.t. additional tree policies and
environment configurations which can be defined separately.

This patch shouldn't change functionality of the explorer, except for
two points which you need to be aware of:

 - The `monte_carlo` option in the config is removed, and the only
   behavior is now that of `monte_carlo = True`.

 - Before, in TAG, we were keeping at most `threshold` evaluations
   across all children of a node; now, we are keeping at most
   `threshold` evaluations per node and collecting the global
   `threshold` highest before selecting a child.  This more faithfully
   implements the original algorithm in the presence of cuts, although
   it shouldn't matter much in practice (branches which are cut are not
   likely to contain good evaluations anyways).

 - Before, in TAG, we were using the number of children before applying
   the cut in the formula.  Instead, they are now ignored, and only the
   number of children after cut is used.  This shouldn't matter since
   this is only a multiplicative factor in the formula, but more closely
   matches the actual tree structure.

The main architectural changes in this patch are as follow:

 - `Children` is renamed to `Node`, as it represents an internal node of
   the tree (arguably, `InternalNode` would be more appropriate).

 - A new `Edge` structure is created for each child of a `Node`.  The
   evaluation statistics are stored on the edge, instead of being in a
   separate vector.

 - `SubTree` and `DescendState` are merged.  This is made easy by the
   removal of the `monte_carlo` option.  This also has the nice effect to
   remove the confusing `descend_noexpand` method.

 - A new `Env` structure is introduced to hold information about the
   environment we are exploring.  This is a wrapper around the context
   and provide a clear hook point to modify the way we explore the
   environment, such as changing the order in which actions are taken, or
   selecting a flat order.

 - A `TreePolicy` trait is introduced.  This contains information about
   how to pick children in the tree (where we have statistics), and how to
   backpropagate evaluation information.  Currently only the existing
   algorithms (random-based and TAG) are implemented, but this can
   easily be extended to handle more classic UCT policies and hybrid
   policies with transversal models.

telamon/
 * src/explorer/bandit_arm.rs:  Refactored to be more generic.
 * src/explorer/local_selection.rs: Make `pick_index` and
`pick_candidate` methods of `NewNodeOrder` instead of free functions.
 * src/explorer/mod.rs:  Give a policy to the tree
 * src/explorer/config.rs:  Remove monte_carlo option

telamon-kernels/
 * benches/cuda_variance.rs:  Use `NewNodeOrder.pick_index`
 * src/kernel.rs:  Use `NewNodeOrder.pick_index`

### DIFF
--- a/src/explorer/bandit_arm.rs
+++ b/src/explorer/bandit_arm.rs
@@ -275,10 +275,10 @@ enum SubTree<'a, P: TreePolicy> {
 
 impl<'a, P: TreePolicy> From<Node<'a, P>> for SubTree<'a, P> {
     fn from(node: Node<'a, P>) -> Self {
-        if node.bound().is_some() {
-            SubTree::InternalNode(Arc::new(node))
-        } else {
+        if node.is_deadend() {
             SubTree::Empty
+        } else {
+            SubTree::InternalNode(Arc::new(node))
         }
     }
 }
@@ -340,7 +340,7 @@ impl<'a, P: TreePolicy> Edge<'a, P> {
         *unwrap!(self.bound.read())
     }
 
-    /// Trims the branch if it has with an evaluation time guaranteed to be worse than
+    /// Trims the branch if it has an evaluation time guaranteed to be worse than
     /// `cut`. Returns the childrens to trim if any,
     fn trim(&self, cut: f64) -> Option<Arc<RwLock<SubTree<'a, P>>>> {
         if self.bound() >= cut {
@@ -364,16 +364,18 @@ impl<'a, P: TreePolicy> Edge<'a, P> {
                         return SubTree::InternalNode(Arc::clone(node))
                     }
                     SubTree::Leaf(_) => {
-                        // Need write access
+                        // Need write access, see below
                     }
                 }
             }
 
-            // Some times we do need write access to expand a leaf candidate... but some other
-            // thread could beat us to the punch and expand it before, so we loop.
+            // Some times we do need write acces to expand a leaf candidate... but another thread
+            // could beat us to the punch and expand it before us, in which case we won't see a
+            // leaf below and loop again (where the first read access should not fail).
             {
                 let dst = &mut *unwrap!(self.dst.write());
                 if let SubTree::Leaf(_) = dst {
+                    // We got write access to the leaf; expand it.
                     if let SubTree::Leaf(candidate) =
                         std::mem::replace(dst, SubTree::Empty)
                     {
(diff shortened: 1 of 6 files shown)


src/explorer/bandit_arm.rs
```diff
@@ -229,289 +254,292 @@ impl<'a, 'b> Store<'a> for Tree<'a, 'b> {
...
-                    if let Some(bound) = children.bound() {
-                        let children = Arc::new(RwLock::new(children));
-                        *self = SubTree::InternalNode(children.clone(), bound);
-                        DescendState::InternalNode(children, true)
+
+            // Some times we do need write access to expand a leaf candidate... but some other
+            // thread could beat us to the punch and expand it before, so we loop.
+            {
+                let dst = &mut *unwrap!(self.dst.write());
+                if let SubTree::Leaf(_) = dst {
```
nicoTolly: It took me some time to understand this code, so it could be worth to expand the comment a bit... So, the call for write is blocking and wait for the release of the lock if it was taken. If it was indeed taken (and we waited a bit) there is a good chance that the expanding was done for us, then dst is no longer a Leaf and we can loop and use it in read-only mode. If dst is indeed a leaf, we have to expand it ourselves. Comment could explain briefly "Dst could not be a leaf anymore if another thread expanded it while we waited for the lock. In this case, we loop"

src/explorer/bandit_arm.rs
```diff
@@ -229,289 +254,292 @@ impl<'a, 'b> Store<'a> for Tree<'a, 'b> {
...
+                    if let SubTree::Leaf(candidate) =
+                        std::mem::replace(dst, SubTree::Empty)
+                    {
+                        let choice = env.list_actions(&candidate);
+                        if let Some(choice) = choice {
+                            let node = Node::from_candidates(
+                                env.apply_choice(&candidate, choice),
+                            );
+                            if node.bound().is_some() {
```
nicoTolly: Couldn't we have something like "is_deadend()" instead of this test on the bound ? It would be more intuitive

src/explorer/bandit_arm.rs
Elarnon: That's what I was trying to say with "some other thread could beat us to the punch and expand it" :)  Will clarify.

src/explorer/bandit_arm.rs
Elarnon: Yeah; I reused the logic from the previous code.  I will create a helper function.

Elarnon:
@nicoTolly are we good on that one?

src/explorer/bandit_arm.rs
```diff
@@ -229,289 +259,306 @@ impl<'a, 'b> Store<'a> for Tree<'a, 'b> {
...
         .iter()
         .enumerate()
-        .filter(|&(i, _)| children[i].bound() < cut)
-        .map(|(ind, x)| (ind, heval(config, x.0.len(), x.1, nb_tested, rewards.len())))
-        .max_by(|x1, x2| cmp_f64(x1.1, x2.1))
-        .map(|(ind, _)| ind)
+        .filter(|(_, edge)| edge.bound() < cut);
+
+    // Compute the threshold to use so that we only have `config.threshold` children
+    let threshold = {
```
nicoTolly: Could we find two different names for the two threshold in this function ? I would keep the name 'threshold' for the f64 variable (as it gives the right intuition) and rename the other one (the argument). Num_children_to_consider ? A bit long. Also it would be better to remain consistent with the name chosen in config (Maybe rename this one too) ?

Elarnon: Renamed the config one to `topk` as per offline discussion!

nicoTolly:
bors retry

PR #177, opened 2018-12-01
[C API] Use repr(transparent) for newtype wrappers

Elarnon:
This patch uses `#[repr(transparent)]` instead of `#[repr(C)]` for
newtype wrappers.   This allows `cbindgen` to generate `typedefs`
instead of single-field structs, which are very non idiomatic in C.

In addition, this adds `#[repr(transparent)]` to `VarId` to make the C
and Python APIs work again.

### DIFF
--- a/telamon-capi/cbindgen.toml
+++ b/telamon-capi/cbindgen.toml
@@ -2,7 +2,7 @@ autogen_warning = """/* DO NOT MODIFY THIS MANUALLY!
  * This file has been generated usin cbindgen.
  *
  * To generate this file:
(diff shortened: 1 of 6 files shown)


telamon-capi/cbindgen.toml
```diff
@@ -2,7 +2,7 @@ autogen_warning = """/* DO NOT MODIFY THIS MANUALLY!
  * This file has been generated usin cbindgen.
  *
  * To generate this file:
- *  1. Get the latest cbindgen using `cargo install --force cbindgen --git https://github.com/Elarnon/cbindgen`
+ *  1. Get the latest cbindgen using `cargo install --force cbindgen`
```
ulysseB: cool :)

PR #178, opened 2018-12-01
Add an `expandconfig` utility to add default values

Elarnon:
This patch adds an `expandconfig` binary which is able to load a
configuration file, and write it back (in either TOML or YAML format)
with the default values expanded.  This is useful in a variety of
scenarios:

 - Check that a configuration file is valid with:

    expandconfig /path/to/config.toml

 - See default field values and structure with:

    expandconfig

 - See all available fields with (YAML displays fields with a default
   `None` value while TOML skips them):

    expandconfig -f yaml


src/bin/expandconfig.rs
```diff
@@ -0,0 +1,113 @@
...
+        program
+    );
+    print!("{}", opts.usage(&brief));
+}
+
+fn main() {
+    let args: Vec<String> = env::args().collect();
+    let program = args[0].clone();
+
+    let mut opts = Options::new();
```
ulysseB: Is their a reason why you didn't use `structopt` as you did in treesize ? I found `structop` much nicer to write/read.

Cargo.toml
```diff
@@ -19,6 +19,9 @@ name = "bench-perf-model"
 path = "tools/bench_perf_model/main.rs"
 required-features = ["cuda"]
 
+[[bin]]
+name = "expandconfig"
```
ulysseB: I don't think you need this entry. Cargo automatically infer files in src/bin are executables.

src/bin/expandconfig.rs
Elarnon: I agree :)

What happened here is that I initially just read from stdin / wrote to stdout without args, then reached for `getopts` to make it a bit nicer since it was already a dependency of `telamon`.  I didn't want to add `structopt`/`clap` dependencies, which are somewhat heavy, just for such a simple utility with two options.  That may be misguided, esp. if we want to use `structopt`/`clap` for other CLI utilities in the future; I'm fine with converting the code if you prefer.

Cargo.toml
Elarnon: Cargo told me otherwise, because we have explicit binaries not in `src/bin`.

> warning: An explicit [[bin]] section is specified in Cargo.toml which currently
> disables Cargo from automatically inferring other binary targets.
> This inference behavior will change in the Rust 2018 edition and the following
> files will be included as a binary target:
> 
> * /home/elarnon/code/telamon/src/bin/expandconfig.rs
>
> This is likely to break cargo build or cargo test as these files may not be
> ready to be compiled as a binary target today. You can future-proof yourself
> and disable this warning by adding `autobins = false` to your [package]
> section. You may also move the files to a location where Cargo would not
> automatically infer them to be a target, such as in subfolders.
>
> For more information on this warning you can consult
> https://github.com/rust-lang/cargo/issues/5330

Elarnon:
@ulysseB `expandconfig` now uses structopt.  I'll try converting the argument parsing code in explorer/config to structopt in a separate PR when I have time.

ulysseB:
Ok. Can I merge this ?

src/bin/expandconfig.rs
```diff
@@ -0,0 +1,180 @@
...
+            Ok(Yaml)
+        } else {
+            Err(ParseFormatError { _priv: () })
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+struct ParseFormatError {
+    _priv: (),
```
ulysseB: why do you need this field ?

Elarnon: I don't need it; it's just to prevent creating the error from somewhere else, which is not actually very useful.  I can remove it.

Elarnon:
bors r=ulysseB

PR #180, opened 2018-12-03
Add UCT tree policy

Elarnon:
This patch adds a new tree policy based on the UCT formula for selecting
children.

telamon/
 * src/explorer/bandit_arm.rs:
        Add UCT policy implementation.  This also slightly changes the `Node`
        and `Edge` structures so that they only depend on the actual statistic
        structures used, not on the policy type.
 * src/explorer/config.rs:  Add UCT configuration options.
 * src/explorer/mod.rs:  Dispatch config to the proper policy.


Elarnon:
Mais.

Elarnon:
bors r=nicoTolly

PR #181, opened 2018-12-03
Improve usability of rustfmt pre-commit hook

Elarnon:
This patch makes the pre-commit hook more useful and forgiving.  In
particular,

 - It no longer silently installs `rustfmt` with `rustup`, instead
   prompting the user to install it

 - It no longer blindly reformats all staged files, instead refusing to
   format files which have both staged and unstaged changes

 - It actually tells the user about reformatted files which should be
   added to the commit, instead of just reformatting and doing nothing.

The hook is also renamed to `pre-commit` from `pre-format` to match the
hook name used by git.  The installation instructions no longer recommend
using a hard link, which can be confusing and oh-so-easy to misuse.


Elarnon:
bors r=ulysseB

Elarnon:
bors retry

PR #184, opened 2018-12-12
Use rpds::List serde feature instead of using a custom wrapper

Elarnon:
Deserialization is slightly space-inefficient since we will have linked
lists instead of vectors; however, it does not matter very much since
this is for log analysis code which can be slow, and it reduces
complexity.

Also fix an issue with pre-commit hook which was not failing when
rustfmt itself fails.


Elarnon:
bors retry

PR #185, opened 2018-12-13
Upgrade formatting with rustfmt stable release

Elarnon:


Elarnon:
This is more ugly than before and changes a bunch of files for nothing, but oh well...

Elarnon:
bors try

PR #186, opened 2018-12-17
Add a CLI tool to extract data from event logs

Elarnon:
The new tool, in `examples/parse_event_log`, allows dumping a CSV of all
the evaluations that were performed during a run, as well as a graphviz
(.dot) graph recapitulating the run.

telamon-utils/
 - src/tfrecord.rs:  Add `.records()` iterator method.

telamon/
 - Cargo.toml:  Add crates for graphviz and csv writing.
 - examples/print_event_log.rs:  Upgrade to the `.records()` API
 - examples/parse_event_log.rs:  Added.

### DIFF
--- a/examples/parse_event_log.rs
+++ b/examples/parse_event_log.rs
@@ -1,3 +1,5 @@
+/// This allows dumping a CSV of all the evaluations that were performed during a run, as well as a
+/// graphviz (.dot) graph recapitulating the run.
 extern crate bincode;
 extern crate csv;
 extern crate dot;
@@ -71,13 +73,13 @@ struct TreeInfo<'a> {
     edges: Vec<(usize, usize, &'a Edge)>,
 }
 
-type Nd<'a> = (usize, &'a &'a Node);
+type Nd<'a> = (usize, &'a Node);
 
 type Ed<'a> = &'a (usize, usize, &'a Edge);
 
 impl<'a> dot::GraphWalk<'a, Nd<'a>, Ed<'a>> for TreeInfo<'a> {
     fn nodes(&'a self) -> dot::Nodes<'a, Nd<'a>> {
-        self.nodes.iter().enumerate().collect()
+        self.nodes.iter().cloned().enumerate().collect()
     }
 
     fn edges(&'a self) -> dot::Edges<'a, Ed<'a>> {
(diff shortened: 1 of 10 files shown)


Elarnon:
bors try

Elarnon:
Forgot to add `Hash` implementations for actions; should be good now.

Elarnon:
bors try

examples/parse_event_log.rs
```diff
@@ -0,0 +1,278 @@
+extern crate bincode;
```
ulysseB: Why put the file in example/ instead of src/bin/ ?

examples/parse_event_log.rs
```diff
@@ -0,0 +1,278 @@
...
+}
+
+impl Node {
+    fn info(&self) -> TreeInfo<'_> {
+        let mut worklist = vec![(self, 0)];
+        let mut nodes = vec![];
+        let mut edges = vec![];
+
+        let max_depth = None;
+        let min_evals = Some(1000);
```
ulysseB: Is this configuration variables ? Shouldn't they be arguments of the function ?

examples/parse_event_log.rs
```diff
@@ -0,0 +1,278 @@
...
+            edges: edges
+                .into_iter()
+                .map(|(from, to, info)| (nodeindex[&from], nodeindex[&to], info))
+                .collect(),
+        }
+    }
+}
+
+fn dig<II>(children: &mut HashMap<ActionEx, Edge>, actions: II, eval: f64, id: usize)
+where
```
ulysseB: Where does the name `dig` comes from ? I am not sure what the function does from the signature. Can you add a comment ?

examples/parse_event_log.rs
```diff
@@ -0,0 +1,278 @@
...
+        children: Default::default(),
+        evaluations: vec![],
+        id: 0,
+        tag: None,
+    };
+
+    let mut evals = Vec::new();
+
+    for (id, record_bytes) in f.records().enumerate() {
+        match bincode::deserialize(&record_bytes?).unwrap() {
```
ulysseB: you can replace the match with a let.

examples/parse_event_log.rs
Elarnon: It walks the tree following the actions in `actions`, recording the evaluation along the way.  `insert` would probably be a better name; I will change it.

examples/print_event_log.rs
```diff
@@ -9,30 +10,15 @@ use utils::tfrecord::{ReadError, RecordReader};
...
-                    if error.kind() == io::ErrorKind::UnexpectedEof
-                        && offset == f.seek(io::SeekFrom::Current(0))?
-                    {
-                        return Ok(());
-                    }
-                }
-                return Err(err);
-            }
+    for record_bytes in (&mut f).records() {
+        match bincode::deserialize(&record_bytes?).unwrap() {
```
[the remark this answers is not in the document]
ulysseB: Ok I didn't realized it was an emum.

examples/parse_event_log.rs
ulysseB: Ok. Not sure a single binary adds much to the compilation time tough. I think It would be cleaner to have it in bin/.

Ideally, all the exploration-related stuff would be in a separate crate.

Elarnon:
bors r=ulysseB

PR #189, opened 2019-01-22
[bandit] Make UCT algorithm customisable

Elarnon:
In order to facilitate performing experiments, this makes the UCT-based
algorithm customizable in term of optimization formula (UCT and pUCT
currently supported), value transformation (execution time vs execution
speed vs log time), sample reduction (max vs mean) and scale.

The UCT and TAG policy are also rewritten to ensure we only read once
node data that can be modified by other threads.

### DIFF
--- a/src/explorer/bandit_arm.rs
+++ b/src/explorer/bandit_arm.rs
@@ -462,13 +462,84 @@ impl TreePolicy for NewNodeOrder {
 /// actually have a cost and 2) the scale is wrong (evaluations are in the e6+ range but we do as
 /// if they were in 0-1).
 pub struct UCTPolicy {
-    factor: f64,
+    exploration_constant: f64,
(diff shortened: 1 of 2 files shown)


nicoTolly:
Now that we have two different policies, a concise high-level description of what these policies do would be nice. It could be either comments near both structure definitions, something in the config file (I personnaly don't think they belong there) or something in the README, with appropriate reference to papers and "See algorithm description in README" in code. Otherwise everything seems fine :)

Elarnon:
@nicoTolly I added some descriptions for the available algorithms, does this look OK to you?  I don't really want to go into *too much* details/explanations as I think this is not the place for it, but there are references and search terms if people want to understand this better.

Elarnon:
bors retry

PR #193, opened 2019-02-08
Improve precision of event log data

Elarnon:
This patch improves the event log format to incorporate additional
information useful for visualization.  The following additional
information is now logged:

 - The thread which found an implementation
 - The depth in the explicit tree when the implementation was found,
   which allows separating between in-tree and rollout actions
 - The performance model bound for the final implementation.  Bounds for
   intermediate nodes are not available (yet).
 - The cut value at the time the implementation was found (this is the
   current best implementation *at the time the descent started from the
   root*).
 - The time at which the implementation was found.
 - The time at which the evaluation was finished and backpropagation
   started.

In addition, dead ends are now logged as well, with the following
information:

 - The depth in the explicit tree when the dead end was found, which
   allows separating between in-tree and rollout actions (if found
   during rollout)
 - The performance bound on the final candidate (if found during
   rollout)
 - The cut value which was used during the descent (if found during
   rollout)
 - The time at which the dead end was found
 - The thread which found the dead end

### DIFF
--- a/src/explorer/bandit_arm.rs
+++ b/src/explorer/bandit_arm.rs
@@ -84,10 +84,65 @@ impl Default for TreeStats {
 }
 
 #[derive(Serialize, Deserialize)]
+pub enum DeadEndSource {
+    /// Dead-end encountered in the tree
+    Tree {
+        /// List of actions defining the dead-end candidate
+        actions: List<choice::ActionEx>,
+    },
+    /// Dead-end encountered in the rollout phase
+    Rollout {
+        /// List of actions defining the dead-end candidate
+        actions: List<choice::ActionEx>,
+        /// Depth in the tree.  The remaining actions were selected during rollout.
+        depth: usize,
+        /// Performance model bound
+        bound: f64,
+        /// Current cut value
+        cut: f64,
+    },
+}
+
+/// The possible tree events.
+/// WARNING:  Changing the enums *will break* any pre-existing eventlog files.  Adding new cases
+/// *at the end only* is safe.
+#[derive(Serialize, Deserialize)]
 pub enum TreeEvent {
     Evaluation {
         actions: List<choice::ActionEx>,
         score: f64,
+    },
+
+    /// A fully-specified implementation was found and evaluated
+    EvaluationV2 {
+        /// List of actions defining the implementation
+        actions: List<choice::ActionEx>,
+        /// Depth in the tree.  The remaining actions were selected during rollout.
+        depth: usize,
+        /// Execution time
+        score: f64,
+        /// Performance model lower bound
+        bound: f64,
+        /// Cut value when the implementation was found.  This is the best implementation at the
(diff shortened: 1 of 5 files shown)


src/explorer/bandit_arm.rs
```diff
@@ -83,12 +83,67 @@ impl Default for TreeStats {
...
+        /// Performance model lower bound
+        bound: f64,
+        /// Cut value when the implementation was found.  This is the best implementation at the
+        /// time the descent started from the root, as threads only synchronize the cut at the
+        /// root.
+        cut: f64,
+        /// Time at which the implementation was found
+        search_end_time: f64,
+        /// Time at which the evaluation finished
+        evaluation_end_time: f64,
```
andidr: It's a bit strange that the end is also logged with the same thread_id, although the evaluation is carried out by somebody else.

src/explorer/bandit_arm.rs
```diff
@@ -83,12 +83,67 @@ impl Default for TreeStats {
...
+        /// Execution time
+        score: f64,
+        /// Performance model lower bound
+        bound: f64,
+        /// Cut value when the implementation was found.  This is the best implementation at the
+        /// time the descent started from the root, as threads only synchronize the cut at the
+        /// root.
+        cut: f64,
+        /// Time at which the implementation was found
+        search_end_time: f64,
```
andidr: Maybe just call this "time" to be coherent with DeadEnd

src/explorer/bandit_arm.rs
Elarnon: There is a single search thread anyways.  I am not sure where else to attach this information (I am not sure it is actually useful, either, to be honest) but I am happy to take suggestions.

src/explorer/bandit_arm.rs
Elarnon: I'd rather rename `time` to `search_end_time` in `DeadEnd`.

src/explorer/bandit_arm.rs
andidr: Maybe just mention in the field's description that this information comes from a different thread

src/explorer/bandit_arm.rs
andidr: Why seach_**end**_time?

Elarnon: Because this is the time at which the search ended (= we found something, even if in that case this is a dead end).  But I agree this is confusing terminology (one implementation vs the whole procedure) so let's do `descent_end_time` instead.

andidr: What about discovery_time (time of discovery of this candidate in the candidate tree)?

Elarnon: `discovery_time` doesn't work very well for dead-ends either, but OK, I'll do that.

Elarnon:
bors r=andidr

PR #194, opened 2019-02-08
Actually stop exploration when a deadend is found in tree

Elarnon:
Due to changes in #193 being incorrectly scoped, in case we found a
deadend in the tree that was concurrently cut by another thread, we
would end up in a tight loop of ignoring it repeatedly.

This was causing the timeout issue from #192.  The issue actually
couldn't happen in the master branch, but was coming from a separate
branch from which #193 was extracted and the fix for #192 backported,
hence the confusion.

This patch removes the fix from #192 which is no longer useful (and was
actually not effective).  Sorry for the noise!

### DIFF
--- a/src/explorer/bandit_arm.rs
+++ b/src/explorer/bandit_arm.rs
@@ -160,8 +160,6 @@ pub struct Tree<'a, 'b, P: TreePolicy> {
     /// Time at which the `Tree` structure was created.  This is a reasonably good approximation of
     /// the time at which the search was started.
     start_time: std::time::Instant,
-    /// Optional timeout after which the search will stop.
-    timeout: Option<std::time::Duration>,
 }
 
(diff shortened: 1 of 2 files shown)


src/explorer/bandit_arm.rs
```diff
@@ -298,27 +317,6 @@ where
                 return None;
             }
 
-            // TODO: We now have two places where we are checking the timeout: here in the tree
-            // search, and in the generic monitor infrastructure.  Historically, we were only
-            // checking in the generic monitor infrastructure to stop evaluating implementations
-            // once the timeout is reached; however, the tree search can spend a long time finding
-            // only deadends but no implementations.  In that case the search would go on forever
```
andidr: In that case,

Elarnon:
bors r=andidr

PR #195, opened 2019-02-08
Compactify the debug log format for readability

Elarnon:
- The `Action(..)` wrapper is not rendered anymore
 - Dimension IDs are represented using `$n` instead of `DimId(n)`
 - Instruction IDs are represented using `@n` instead of `InstId(n)`
 - Statements are naked, i.e. instructions are represented as `@n`
   instead of `Inst(InstId(n))` and dimensions as `$n` instead of
   `Dim(DimId(n))`.
 - Numeric sets are represented as `NS { 2, 4 }` instead of
   `NumericSet { enabled_values: 6 }`.  Ideally we would want the actual
   values here, but that would require foregoing the `Debug` trait for a
   custom one since this requires knowing the universe at printing time.

### DIFF
--- a/telamon-gen/src/print/runtime/integer_set.rs
+++ b/telamon-gen/src/print/runtime/integer_set.rs
@@ -7,10 +7,20 @@ use quote::quote;
 /// Defines the `NumericSet` type.
 pub fn get() -> TokenStream {
     quote! {
-        #[derive(Copy, Clone, Eq, PartialEq, Hash, Debug, Serialize, Deserialize)]
+        #[derive(Copy, Clone, Eq, PartialEq, Hash, Serialize, Deserialize)]
         #[repr(C)]
         pub struct NumericSet {
             enabled_values: u16
+        }
+
+        impl std::fmt::Debug for NumericSet {
+            fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+                write!(f, "NS{{")?;
+                for (i, bit) in self.list().enumerate() {
+                    write!(f, "{}{}", if i == 0 { "" } else { ", " } , bit.enabled_values)?;
+                }
+                write!(f, "}}")
+            }
         }
 
         #[allow(dead_code)]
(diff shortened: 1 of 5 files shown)


telamon-gen/src/print/runtime/integer_set.rs
```diff
@@ -7,12 +7,22 @@ use quote::quote;
...
         #[repr(C)]
         pub struct NumericSet {
             enabled_values: u16
         }
 
+        impl std::fmt::Debug for NumericSet {
+            fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+                write!(f, "NS{{")?;
+                for (i, bit) in self.list().enumerate() {
+                    write!(f, "{}{}", if i == 0 { "" } else { ", " } , bit.enabled_values)?;
```
ulysseB: you can use '.format()' from itertools to print comma-separated lists.

Elarnon: Right; I forgot we have a dependency on `itertools` already, so no need to worry about adding it.  I'll change it.

Elarnon:
- I'd rather not use `d` and `i` prefix because that would make the names valid Rust identifiers.  `@` and `$` symbols won't occur naturally and are easier to find quickly while scanning a line.  From my experience the meaning of `@` and `$` can be picked up very quickly.
 - The reason for the `NS` prefix is because what is displayed is not the values but rather the indices into the universe (`NS { 2 }` can have different values depending on the universe it is associated with).  I'd rather `{ 2 }` means "the set containing the value 2", which needs some more infrastructure to actually be printed properly, so I used `NS { 2 }` to reduce confusion without decreasing readability too much.

ulysseB:
Ok for 'NS'. Still not totally convinced by '$' and '@'. What is the problem with them being valid identifiers ? Alternatively, can we use something like 'i.3'/'d.3' (or 'i@3'/'d@3' if the dot is a problem) ?

I wont veto if you still think '$' and '@' are the best.

Elarnon:
The main point about using sigils is that they are easier to find quickly when looking at a large blob of text (because they won't occur inside regular identifiers, and have large-ish unusual shapes).  The problem with being valid identifiers is that it restricts what we can do with the log format -- if we use things that are not valid identifiers, we can use them as-is in the same place as we would use an identifier (e.g. action name) without causing confusion and "remove the need to display one layer of enum".  See for instance what this patch does with `StmtId` : since `DimId` and `InstId` are already self-describing, there is no need to add extra info for the `StmtId::Dim` and `StmtId::Inst` constructors.

Elarnon:
bors r=ulysseB

PR #198, opened 2019-02-21
Move the Cuda Backackend to a Separate Crate

ulysseB:
This PR moves the cuda backend into a separate crate to speed-up compilation.


ulysseB:
bors try

ulysseB:
bors try

ulysseB:
try

ulysseB:
bors try

Cargo.toml
```diff
@@ -89,30 +73,29 @@ bincode = "1.0"
...
-[dependencies.tokio-timer]
-git = "https://github.com/ulysseB/tokio-timer.git"
+tokio-timer = { git = "https://github.com/ulysseB/tokio-timer.git" }
 
 [dev-dependencies]
 gcc = "0.3.54"
 tempdir = "0.3.7"
 criterion = "0.2.4"
 jemalloc-ctl = "0.1"
+telamon-cuda = { path = "backend/cuda", default-features = false }
```
Elarnon: Does the `telamon` crate actually use the `telamon-cuda` crate?  At a glance (and with my understanding of the code) it is only the dependent crates (`telamon-kernels`, `telamon-capi`) that use it.

ulysseB: That's only for the tests/examples. I wasn't sure if I should move them to backend/cuda.

ulysseB:
bors try

Cargo.toml
Elarnon: I would say yes as it simplifies the dependency structure.  It feels a bit weird to have the examples (tests less so maybe) in backends/cuda but eventually we probably want to do something like rename the current `telamon` crate to `telamon-core` and have a new `telamon` crate which re-exports relevant stuff from the `telamon-*` crates and contains global examples/tests?

backend/cuda/src/api/fake.rs
```diff
@@ -14,18 +15,9 @@ pub trait Argument: Send + Sync {
...
-{
-    fn as_size(&self) -> Option<u32> {
-        device::ScalarArgument::as_size(self)
-    }
-}
-
 impl Argument for Box<dyn device::ScalarArgument> {
     fn as_size(&self) -> Option<u32> {
-        device::ScalarArgument::as_size(self.as_ref())
+        (self.deref()).as_size()
```
Elarnon: Nit: I think idiomatic rust would be `(*self).as_size()`

backend/cuda/src/context.rs
```diff
@@ -106,12 +103,14 @@ impl<'a> Context<'a> {
         // earlier and make tests to know when (for example, measure the MAX delta between
         // min and median with N outliers).
         let runtimes = (0..num_evals).map(|_| thunk.execute());
-        let runtimes_by_value = process_results(runtimes, |iter| iter.sorted())?;
+        let runtimes_by_value =
+            (process_results(runtimes, |iter| iter.sorted())?).collect_vec();
```
Elarnon: As far as I know you can write `process_results(runtimes, |iter| iter.sorted())?.collect_vec()` without the extra parens.

Elarnon:
r=me with the few nits addressed

Cargo.toml
ulysseB: If that is fine with you, I am going to let it like this and move telamon to telamon-core in a separate commit.

Elarnon: :+1:

ulysseB:
bors r=elarnon

PR #200, opened 2019-02-28
Rewrite MCTS API with a shadow tree

Elarnon:
This patch rewrites the MCTS exploration algorithm to use a "shadow
tree", recording all the information which is known about the nodes that
have been encountered thus far -- both in the "explicit tree" and in the
rollout portions.

The explicit tree records information that was previously learnt, such
as whether some children are dead or their bound, which helps both in
reducing the number of propagations required and eliminates the issue of
"discovering" the same dead-ends multiple times.

The new infrastructure was also written with better support for logging
and tracing in mind, allowing the generation of finer-grain event
traces.  The new traces contain more precise information about the
decisions made during the descent (which child are selected and why,
etc.).

Finally, this patch adds an option for locally backtracking dead-ends
until a live node is found.  This may help with discarding parts of the
tree that are ill-understood by the performance model, although
preliminary experiments show that this is still too expensive.  However,
the amount of backtracking needed to find a live candidate in a specific
portion of the tree should prove to be a good evaluator of how good the
performance model is in that portion, in turn turning to be helpful in
evaluating how good a choice ordering is.

Changes:
 - src/explorer/:
	mcts.rs:  Added.
	bandit_arm.rs:  Support for backtracking deadends.
	local_selection.rs:  Backtracking rollout.
	mod.rs:  Use the new MCTS API.
	config.rs:  MCTS configuration and backtrack_deadends option.
 - src/model/:
	hw_pressure.rs:  Bound is now serializable.
 - telamon-py/:
	setup.py:  Support for out-of-tree builds.


src/explorer/mcts.rs
```diff
@@ -0,0 +1,1712 @@
...
+
+/// Helper structure to manipulate a tree.
+///
+/// Despite the name, the `Tree` does not hold pointers to actual nodes.
+pub struct Tree<'a> {
+    /// The environment to use for computing children.
+    env: Env<'a>,
+    /// Sequential counter for ID attribution.
+    id_counter: &'a AtomicUsize,
+    /// Channel to send trace events to.
```
andidr: Channel to send **log** events to?

src/explorer/mcts.rs
```diff
@@ -0,0 +1,1712 @@
...
+        parent: Option<(NodeId, usize)>,
+        /// All available children for this node.  This includes *all* actions ever considered,
+        /// even those that end up being forbidden due to constraint propagation.
+        children: Vec<Action>,
+        /// Explained bound from the performance model.  `None` if the node was killed due to
+        /// constraint propagation, in which case we can't run the performance model.
+        bound: Option<Bound>,
+    },
+
+    /// Path in the tree followed by a specific thread.  Starts at the root of the tree.
```
andidr: "Path" suggests that the node selected at a step is a neighbor of the previously selected node in the tree. However, a SelectNode event could select any node.

Elarnon: Chagned to sequence of actions

src/explorer/mcts.rs
Elarnon: sure

Elarnon:
Oh almost forgot bors try

Elarnon:
bors try

PR #203, opened 2019-03-04
Make NodeId newtype easier to use

Elarnon:
This also exports the `explorer::mcts` module from `telamon::explorer`.


Elarnon:
bors r=me

PR #204, opened 2019-03-05
Make new MCTS implementation the default

Elarnon:
This patch makes the new MCTS implementation (in explorer::mcts) the
default over the previous explorer::bandit_arm.  explorer::bandit_arm is
kept for the moment for backwards compatibility, although the behavior
of the two implementations should be fairly similar.

The tests are now run using the new MCTS implementation, which allowed
to spot a couple of bugs -- namely, that the new implementation was not
properly detecting tree exhaustion and was not stopping in that case.
The patch also fixes that issue by marking nodes that were already
evaluated as dead, ensuring they won't be evaluated again in the future.
A little bit of thought may be needed to understand what that entails on
the theoretical "guarantees" of the MCTS algorithm.

### DIFF
--- a/src/explorer/mcts.rs
+++ b/src/explorer/mcts.rs
@@ -87,8 +87,8 @@ struct NodeInner<'c, N, E> {
     /// and was never live.
     bound: Option<Box<Bound>>,
 
-    /// If the node is dead, the reason for which it was killed.
-    cause_of_death: RwLock<Option<CauseOfDeath>>,
+    /// Whether the node is dead.
+    dead: AtomicBool,
 
     /// Whether the node was expanded.
     ///
@@ -147,13 +147,20 @@ impl<'c, N, E> Node<'c, N, E> {
 
     /// Returns whether the node is still live.
     pub fn is_live(&self) -> bool {
-        self.inner.bound.is_some()
-            && self
-                .inner
-                .cause_of_death
-                .read()
-                .expect("cause_of_death: poisoned")
-                .is_none()
+        self.inner.bound.is_some() && !self.inner.dead.load(Ordering::Acquire)
+    }
+
+    fn kill(&self) {
+        // TODO: Do not overwrite cause if there already is one?
+        self.inner.dead.store(true, Ordering::Release);
+
+        // Clear the candidate if there is one.  Always do it after killing the node so that if
+        // somebody tries to take the candidate and it fails, they will always see the node as dead
+        // afterwards.
+        self.take_candidate();
+
+        // TODO: node.data.clear()
+        // TODO: for child in node.children { child.data.clear() }
     }
 
     /// Returns whether the node is an implementation, i.e. represents a fully-specified candidate.
@@ -461,7 +468,7 @@ impl<'a> Tree<'a> {
     {
         assert!(parent.is_some() || candidate.is_some());
 
-        let (children, bound, cause_of_death);
+        let (children, bound);
         if let Some(candidate) = candidate {
             children = self
                 .env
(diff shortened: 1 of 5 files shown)


Elarnon:
bors try

src/explorer/mcts.rs
```diff
@@ -976,7 +977,14 @@ where
...
 
         assert!(cursor.node.is_expanded());
-        assert!(!cursor.node.is_implementation());
+
+        // If we reached an implementation in-tree and we did not expand it, it must have been
+        // evaluated by another thread -- backtrack.
+        if cursor.node.is_implementation() {
+            // We don't assert! that the node is dead because the other thread could be currently
+            // evaluating it.
+            return Err(Error::DeadEnd(cursor));
```
ulysseB: Why is this considered as a dead-end ? Wont it affect the statistics ?

src/explorer/mcts.rs
```diff
@@ -147,13 +147,20 @@ impl<'c, N, E> Node<'c, N, E> {
...
-                .cause_of_death
-                .read()
-                .expect("cause_of_death: poisoned")
-                .is_none()
+        self.inner.bound.is_some() && !self.inner.dead.load(Ordering::Acquire)
+    }
+
+    fn kill(&self) {
+        // TODO: Do not overwrite cause if there already is one?
+        self.inner.dead.store(true, Ordering::Release);
```
ulysseB: I would tend to only use SeqCst unless we really need to use another memory Ordering. This makes sure we do not violate invariants in future modifications.

src/explorer/mcts.rs
```diff
@@ -147,13 +147,20 @@ impl<'c, N, E> Node<'c, N, E> {
...
-                .inner
-                .cause_of_death
-                .read()
-                .expect("cause_of_death: poisoned")
-                .is_none()
+        self.inner.bound.is_some() && !self.inner.dead.load(Ordering::Acquire)
+    }
+
+    fn kill(&self) {
+        // TODO: Do not overwrite cause if there already is one?
```
ulysseB: Is this still valid ? I thought the cause field was removed.

src/explorer/mcts.rs
```diff
@@ -147,13 +147,20 @@ impl<'c, N, E> Node<'c, N, E> {
...
+        self.inner.dead.store(true, Ordering::Release);
+
+        // Clear the candidate if there is one.  Always do it after killing the node so that if
+        // somebody tries to take the candidate and it fails, they will always see the node as dead
+        // afterwards.
+        self.take_candidate();
+
+        // TODO: node.data.clear()
+        // TODO: for child in node.children { child.data.clear() }
```
ulysseB: Are theses todos still valid ?

Elarnon: Yes, as optimisation -- to remove data for nodes which have been killed.  We can keep the data in memory for now.

src/explorer/mcts.rs
Elarnon: Indeed, removing

src/explorer/mcts.rs
Elarnon: This should actually be `Ordering::Relaxed` -- we don't use this for any type of synchronisation

Elarnon: Actually we may want to so let's do SeqCst.

src/explorer/mcts.rs
Elarnon: It will affect the statistics, but this should never happen in practice (so I guess there should be a big fat `warn!` logging here).  In fact, it should only ever happen if the *root* is an implementation (and multiple threads are concurrently trying to evaluate it), since the case should be handled when picking the children otherwise.

I think we will agree it is not a very interesting case -- however to make sure that it is the case I will rather move the check to the `explore` function instead, adding a precondition that `select_intree` must only ever be called on non-implementation, expanded nodes.

Elarnon:
We are keeping two implementations in parallel because I want the old implementation to still be around if there turns out to be issues with the new one (which uses the shadow tree we talked about).  Arguably we can dig it up from the git history, so maybe I should remove it.

Elarnon:
bors r=ulysseB

Elarnon:
bors cancel

Elarnon:
bors r=ulysseB r=nicoTolly

PR #205, opened 2019-03-08
Factorize and modernize eventlog reading code

Elarnon:
This patch simplifies the handling of eventlog reading by:

 - Introducing an `eventlog` module which takes care of opening the
   eventlog files properly wrt compression, and uses it when opening
   eventlog files.

 - Refactors the tfrecord reading code to expose an interface similar to
   that of "wrapper" formats such as the `csv` crate.  More
   specifically, the tfrecord reader and writer traits are converted to
   wrapping structures.


Elarnon:
bors r=andidr

PR #206, opened 2019-03-08
Properly log dead-ends in the trace

Elarnon:
When a trace ends up in a dead-end, we were discarding the trace, while
the intent was to log it beforehand -- which this patch does.


Elarnon:
bors r=andidr

PR #207, opened 2019-03-08
Log the probability distributions used for selecting children

Elarnon:
This patch adds to the log events the information about *how* a child
node was selected.  For now, this means adding either the probability
distribution (for random picking) or the scores (for deterministic
picking) associated with each node; in the future, we may log more
precise information such as the raw state of the policy -- from which
the aforementioned values can be computed.


Elarnon:
Properly set the `start_time` to the beginning of the exploration instead of the start of the current descent.

Elarnon:
bors r=andidr

Elarnon:
bors retry

PR #209, opened 2019-03-11
Add a round-robin baseline policy

Elarnon:
This adds a baseline policy using round-robin selection, i.e. which is
selecting the child which was visited least at each step.

### DIFF
--- a/src/explorer/mcts.rs
+++ b/src/explorer/mcts.rs
@@ -1453,7 +1453,7 @@ impl UCTPolicy {
     fn value(&self, stats: &UCTStats) -> (f64, usize) {
         use self::config::ValueReduction;
 
-        let num_visits = stats.num_visits();
+        let num_visits = stats.common.num_visits();
 
         let value = match self.value_reduction {
             ValueReduction::Best => stats.best_evaluation(),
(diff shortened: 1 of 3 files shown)


src/explorer/mcts.rs
```diff
@@ -2001,3 +2018,32 @@ impl<'a> IntoIterator for &'a Evaluations {
...
+        &'_ self,
+        _cut: f64,
+        view: &NodeView<'_, 'c, N, CommonStats>,
+    ) -> Option<(EdgeViewIndex, Selector<EdgeIndex>)> {
+        Selector::maximum(
+            view.iter()
+                .map(|(index, edge, _node)| (index, -(edge.data().num_visits() as f64)))
+                .collect(),
+        )
+        .map(|selector| {
```
nicoTolly: I had a hard time understanding that this map is not the map implemented in Selector but the map of Option<Selector> (I finally guessed based on the closure argument being named selector, which induces the type). Maybe a comment clarifying what is happening would be worth here ? "Map Option<Selector<Node>> to Option<Selector<Edge>> "

Elarnon: If this needs a comment everywhere that's not going to work.  I'd rather rename `Selector::maximum` to `Selector::try_maximum` (and `Selector::random` to `Selector::try_random` to be consistent) to show the result is not a `Selector` instance if that's OK with you.

nicoTolly: Ok :)

Elarnon:
bors r=nicoTolly

hmpf.

PR #210, opened 2019-03-12
Log node discovery time and evaluation results time

Elarnon:
This also makes the fields on `Timed` public so that events can be used
outside of the `explorer::mcts` module.


Elarnon:
bors r=andidr

PR #212, opened 2019-03-13
Add code for off-line reconstruction of candidate trees

andidr:
The new module offline_analysis allows for the reconstruction of a candidate
tree from the events of a log file for post-mortem analysis or conversion to a
different log format required by an analysis tool.

The reconstruction of the tree itself and refinement of candidate nodes from
actions on the nodes can be done incrementally and in an interleaved fashion as
log messages are processed. The general outline of this process is:

  let mut tree = CandidateTree::new();
  let mut actions = ThreadActionList::new();

  // Iterate over all messages from the log file
  for ... {
    match message {
      Message::Node { ... } => { tree.extend(...); }

      Message::Trace { ... } => {
        // Iterate over all events of the trace message
        for event ... {
          match event {
            Event::Kill(...) => {
              // Declare candidate as a deadend
              tree.get_node(curr_node_id).borrow_mut().declare_deadend(...);
	      ...
            }

            Event::Implementation(...) => {
              // Update score
              tree.get_node(curr_node_id)
	        .borrow_mut()
		.declare_implementation(...);
       	      ...
            }

            ...
          }
        }
      }

      Message::Evaluation { ... } => {
        ...
        t.get_node(..).borrow_mut().set_score(...);
      }
    }
  }

### DIFF
--- a/src/offline_analysis/mod.rs
+++ b/src/offline_analysis/mod.rs
@@ -1,0 +1,1 @@
+pub mod tree;
--- a/src/offline_analysis/tree.rs
+++ b/src/offline_analysis/tree.rs
@@ -1,0 +1,481 @@
+///! Data structures and function that allow for the recreation of a
+///! candidate tree from a log file
+use crate::explorer::choice::ActionEx as Action;
+use crate::explorer::mcts::{EdgeIndex, NodeId};
+use crate::model::Bound;
+use core::ops::Deref;
+use std::cell::{Ref, RefCell};
+use std::rc::{Rc, Weak};
+use std::time::Duration;
+use utils::FnvHashMap;
+
+/// Outgoing Edge to a child annotated with the action for the
+/// child. If `child` is None, the child node corresponding to the
+/// action has not been computed or hasn't been added to the tree,
+/// yet.
+struct ChildEdge {
+    action: Action,
+    child: Option<Weak<RefCell<CandidateNodeInner>>>,
+}
+
+/// Edge to parent with `child_idx` indicating the index within the
+/// list of children of `parent`.
+pub struct ParentEdge {
+    parent: Weak<RefCell<CandidateNodeInner>>,
+    child_idx: EdgeIndex,
+}
+
+impl From<EdgeIndex> for usize {
+    fn from(v: EdgeIndex) -> Self {
+        usize::from(u16::from(v))
+    }
+}
+
+/// Inner object for nodes of in the resconstructed candidate tree, wrapped by the proxy CandidateNode
+struct CandidateNodeInner {
+    /// Incoming edge from the parent to this node. May be None if
+    /// this is the root node.
+    incoming_edge: Option<ParentEdge>,
+
+    /// Time at which the node was discovered
+    discovery_time: Duration,
+
+    /// Time at which the node was marked as an internal node
+    internal_time: Option<Duration>,
+
+    /// Time at which the node was marked as a rollout node
+    rollout_time: Option<Duration>,
+
+    /// Time at which the node was marked as an implementation
+    implementation_time: Option<Duration>,
+
+    /// Time at which the node was marked as a deadend
+    deadend_time: Option<Duration>,
+
+    /// Bound determined by the performance model
+    bound: Option<Bound>,
+
+    /// List of edges to the node's children with the corresponding
+    /// action.
+    outgoing_edges: Vec<ChildEdge>,
+
+    /// ID of this node
+    id: NodeId,
+
+    /// Score from the evaluation
+    score: Option<f64>,
+}
+
+trait ReplaceDurationIfLower {
+    /// Set duration to d if currently undefined or if d is lower
+    fn replace_if_lower(&mut self, d: Duration);
+}
+
+impl ReplaceDurationIfLower for Option<Duration> {
+    fn replace_if_lower(&mut self, d: Duration) {
+        if self.is_none() || self.unwrap() < d {
+            self.replace(d);
+        }
+    }
+}
+
+/// A node of in the resconstructed candidate tree
+pub struct CandidateNode {
+    inner: Rc<RefCell<CandidateNodeInner>>,
+}
+
+/// Proxy for bound, capturing bound from CandidateNode::bound()
+pub struct BoundGuard<'a> {
+    guard: Ref<'a, CandidateNodeInner>,
+}
+
+impl<'b> Deref for BoundGuard<'b> {
+    type Target = Option<Bound>;
+
+    fn deref(&self) -> &Self::Target {
+        &self.guard.bound
+    }
+}
+
+/// A node in the re-created tree representing a candidate
+impl CandidateNode {
+    /// Returns the time at which the node was discovered
+    pub fn discovery_time(&self) -> Duration {
+        self.inner.borrow().discovery_time
+    }
+
+    /// Returns the time at which the node was marked as an internal
+    /// node
+    pub fn internal_time(&self) -> Option<Duration> {
+        self.inner.borrow().internal_time
+    }
+
+    /// Returns the time at which the node was marked as a rollout
+    /// node
+    pub fn rollout_time(&self) -> Option<Duration> {
+        self.inner.borrow().rollout_time
+    }
+
+    /// Returns the time at which the node was marked as an
+    /// implementation
+    pub fn implementation_time(&self) -> Option<Duration> {
+        self.inner.borrow().implementation_time
+    }
+
+    /// Returns the time at which the node was marked as a deadend
+    pub fn deadend_time(&self) -> Option<Duration> {
+        self.inner.borrow().deadend_time
+    }
+
+    /// Returns the bound from the performance model for this candidate
+    pub fn bound(&self) -> BoundGuard {
+        BoundGuard {
+            guard: self.inner.borrow(),
+        }
+    }
+
+    /// Returns the score from the evaluation of this candidate
+    pub fn score(&self) -> Option<f64> {
+        self.inner.borrow().score
+    }
+
+    /// Returns the parent node or None if this is the root node
+    pub fn parent(&self) -> Option<CandidateNode> {
+        self.inner
+            .borrow()
+            .incoming_edge
+            .as_ref()
+            .map(|edge| CandidateNode {
+                inner: edge.parent.upgrade().unwrap(),
+            })
+    }
+
+    /// Returns true if this node has been explored (i.e., it is a
+    /// rollout node or an internal node)
+    pub fn is_explored(&self) -> bool {
+        self.is_internal_node() || self.is_rollout_node()
+    }
+
+    /// Returns the child at index `child_idx`. This may be None if
+    /// the child hasn't been set before.
+    ///
+    /// # Panics
+    /// Panics if the index is invalid.
+    pub fn child(&self, child_idx: usize) -> Option<CandidateNode> {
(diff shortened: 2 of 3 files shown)


src/offline_analysis/tree.rs
```diff
@@ -0,0 +1,325 @@
...
+
+        assert!(id == 0, "Attempting to add root node with an ID != 0");
+
+        self.add_node_mapping(Rc::clone(&root), id);
+        self.root = Some(root);
+    }
+
+    /// Adds a new mapping from the given `id` to the node `n`. Panics
+    /// if a mapping for the given ID already exists.
+    fn add_node_mapping(&mut self, n: Rc<RefCell<CandidateNode>>, id: u64) {
```
Elarnon: Why do you take the `id` as argument instead of using `n.id()`?

src/offline_analysis/tree.rs
```diff
@@ -0,0 +1,325 @@
...
+
+        if let Some((parent_id, child_idx)) = parent {
+            let child = {
+                let parent = self.node_ids.get(&parent_id).unwrap();
+                new_node
+                    .borrow_mut()
+                    .set_parent(Rc::clone(parent), child_idx);
+
+                let mut parent_node = parent.borrow_mut();
+                parent_node.set_child(child_idx, Rc::clone(&new_node));
```
Elarnon: You should set the parent directly when creating the `new_node` instead.

src/offline_analysis/tree.rs
```diff
@@ -0,0 +1,325 @@
...
+    /// declared the node as an implementation with a lower timestamp.
+    pub fn declare_implementation(&mut self, timestamp: Duration) {
+        if !self.is_implementation() || self.implementation_time.unwrap() > timestamp {
+            self.implementation_time = Some(timestamp);
+        }
+    }
+
+    /// Sets the score from an evaluation
+    pub fn set_score(&mut self, score: f64) {
+        self.score = Some(score);
```
ulysseB: Is it allowed to overwrite a score ? If not, use `assert!(self.score.replace(score).is_none())`.

src/offline_analysis/tree.rs
```diff
@@ -0,0 +1,325 @@
...
+    }
+}
+
+/// A reconstructed tree
+pub struct CandidateTree {
+    /// Virtual root node of the reconstructed tree
+    root: Option<Rc<RefCell<CandidateNode>>>,
```
ulysseB: What does virtual means ?

src/offline_analysis/tree.rs
```diff
@@ -0,0 +1,325 @@
...
+        CandidateTree {
+            root: None,
+            node_ids: HashMap::new(),
+        }
+    }
+
+    /// Returns the root node; panics if no root has been created
+    /// beforehand
+    pub fn get_root(&self) -> Rc<RefCell<CandidateNode>> {
+        Rc::clone(self.root.as_ref().unwrap())
```
ulysseB: You could return a reference instead.

src/offline_analysis/tree.rs
```diff
@@ -0,0 +1,325 @@
...
+    }
+
+    /// Returns the node with the given `id`. Panics if no such node
+    /// exists in the tree.
+    pub fn get_node(&self, id: u64) -> Rc<RefCell<CandidateNode>> {
+        Rc::clone(self.node_ids.get(&id).expect(&format!(
+            "Attempting to retrieve unknown node with id {}",
+            u64::from(id)
+        )))
```
ulysseB: Idem: why copy ?

src/offline_analysis/tree.rs
```diff
@@ -0,0 +1,325 @@
...
+    }
+
+    /// Checks whether a node with the given `id` exists in the tree.
+    pub fn has_node(&self, id: u64) -> bool {
+        self.node_ids.get(&id).is_some()
+    }
+
+    /// Setst he root node to `root` and associates it with the given `id`.
```
ulysseB: typo

Elarnon:
BTW it would be good to provide an API for creating a `CandidateTree` from a `Message` iterator.  The simplest is to provide a `fn parse<II>(ii: II) -> CandidateTree where II: IntoIterator<Item = Message>` method on `CandidateTree`.

src/offline_analysis/tree.rs
```diff
@@ -0,0 +1,325 @@
...
+        &mut self,
+        node_id: u64,
+        discovery_time: Duration,
+        parent: Option<(u64, usize)>,
+        bound: Option<f64>,
+        actions: Vec<Action>,
+    ) -> Rc<RefCell<CandidateNode>> {
+        let num_children = actions.len();
+
+        let new_node = Rc::new(RefCell::new(CandidateNode {
```
ulysseB: You could have a function `CandidateNode::new(..) -> CandiadateNode` that does this. You could also directly set the parent in the child.

src/offline_analysis/tree.rs
```diff
@@ -0,0 +1,325 @@
+///! Data structures and function that allow for the recreation of a
+///! candidate tree from a log file
+use crate::explorer::choice::ActionEx as Action;
+use std::cell::RefCell;
+use std::collections::HashMap;
```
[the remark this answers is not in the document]
Elarnon: As a counterpoint, I really don't like this because `HashMap` is a very standard name, and using `utils::HashMap` (which usually gets imported through `use utils::*`) is generally confusing.

As it turns out, `utils::HashMap` is nothing else than `fnv::FnvHashMap` (from the `fnv` crate, which `telamon-utils` uses), and I would be very much in favour of renaming our use of `HashMap` into this instead (which I think I'll actually do now).

src/offline_analysis/mod.rs
```diff
@@ -0,0 +1 @@
+pub mod tree;
```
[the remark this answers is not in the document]
andidr: yes, there will be another one called "action"

src/offline_analysis/tree.rs
```diff
@@ -0,0 +1,325 @@
...
+
+    /// Actions that correspond to the edges from this node to the
+    /// respective child.
+    children_actions: Vec<Action>,
+
+    /// ID of this node
+    pub id: u64,
+
+    /// Score from the evaluation
+    pub score: Option<f64>,
```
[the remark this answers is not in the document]
andidr: It's the execution time obtained from an evaluation of the candidate

src/offline_analysis/tree.rs
```diff
@@ -0,0 +1,325 @@
...
+    /// Returns true if this node is neither a rollout node nor an
+    /// internal node
+    pub fn is_unexplored(&self) -> bool {
+        !self.is_internal_node() && !self.is_rollout_node()
+    }
+
+    /// Sets the child at index `child_idx` to `child`. Panics if the
+    /// child was already set before or if the index is invalid.
+    fn set_child(&mut self, child_idx: usize, child: Rc<RefCell<CandidateNode>>) {
+        assert!(child_idx < self.children.len(), "Cannot set child with index {}, node was announced to have only {} children.", child_idx, self.children.len());
```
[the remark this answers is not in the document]
andidr: I was expecting rustfmt to do this automatically given the values in rustfmt.toml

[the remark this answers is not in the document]
Elarnon: `rustfmt` gets confused with long strings. (btw `Vec` already does this bound check and I am not 100% convinced the custom error message improves much on the default one)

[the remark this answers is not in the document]
andidr: I prefer getting a specific error message putting the erroneous arguments into context rather some generic panic.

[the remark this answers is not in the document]
ulysseB: In that case, you can merge the error message and the array access: `myarray.get_mut(child_idx).expect("bla")`

src/offline_analysis/tree.rs
```diff
@@ -0,0 +1,497 @@
...
+    /// ID of this node
+    id: NodeId,
+
+    /// Score from the evaluation
+    score: Option<f64>,
+}
+
+trait ReplaceDurationIfLower {
+    /// Set duration to d if currently undefined or if d is lower
+    fn replace_if_lower(&mut self, d: Duration);
```
Elarnon: Nit: Using a trait for this instead of a free-standing function is a bit overkill, but OK.

src/explorer/mcts.rs
```diff
@@ -73,6 +80,26 @@ impl From<EdgeIndex> for u16 {
     }
 }
 
+impl From<EdgeIndex> for usize {
+    fn from(v: EdgeIndex) -> Self {
+        v.0 as usize
+    }
+}
+
+impl AddAssign<u16> for EdgeIndex {
```
Elarnon: Please remove those implementations and manually set create the corresponding edges where needed.  The point of having `EdgeIndex` as a type instead of using `usize` or some such is to make it hard to construct invalid edge indices, so implementing `AddAssign` and `Sub` here somewhat defeat the purpose.

src/offline_analysis/tree.rs
```diff
@@ -0,0 +1,497 @@
...
+
+/// A node of in the resconstructed candidate tree
+pub struct CandidateNode {
+    inner: Rc<RefCell<CandidateNodeInner>>,
+}
+
+/// An iterator to iterate over the children of a CandidateNode
+pub struct CandidateNodeChildrenIterator {
+    parent: CandidateNode,
+    child_idx: EdgeIndex,
```
Elarnon: This should just use an `u16` index and cast it appropriately when calling `.child()` instead.  As said above the point of `EdgeIndex` is to avoid accidentally create wrong values -- we are creating the values by incrementing a counter here anyways, so they don't provide much.

src/offline_analysis/tree.rs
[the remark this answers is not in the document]
ulysseB: can you use utils::FnvHashMap instead ?

src/offline_analysis/tree.rs
```diff
@@ -0,0 +1,503 @@
...
+            .as_ref()
+            .map(|child_weak| CandidateNode {
+                inner: child_weak.upgrade().unwrap(),
+            })
+    }
+
+    /// Returns an interator that allows for iteration over all
+    /// children of this node as Option<CandidateNode>, including
+    /// unexplored children represented by None values.
+    pub fn children(&self) -> CandidateNodeChildrenIterator {
```
ulysseB: An anonymous type for the iterator would have sufficed.
```rust
pub fn children(&self) -> impl Iterator<Item=Option<CandidateNode>> {
    let inner = Rc::clone(self.inner);
    (0..self.num_children()).map(move |i| inner.child(i))
}
```

src/offline_analysis/tree.rs
```diff
@@ -0,0 +1,503 @@
...
+
+        if let Some((parent_id, child_idx)) = parent {
+            let parent_node = self.get_node(parent_id);
+            let uidx = usize::from(child_idx);
+
+            assert!(uidx < parent_node.num_children());
+
+            {
+                let parent_out_edge =
+                    &mut parent_node.inner.borrow_mut().outgoing_edges[uidx];
```
ulysseB: use `outgoing_edges.get(uidx).expect()` instead of the assert

Elarnon:
bors r=ulysseB, r=Elarnon

andidr:
bors retry

andidr:
bors retry

PR #213, opened 2019-03-13
Be explicit about using HashMap *not* being the standard HashMap

Elarnon:


Elarnon:
I think those happen in old files where `rustfmt` is usually not run (because they are not used anymore) but here it was run because they were modified.

Elarnon:
bors r=ulysseB

PR #214, opened 2019-03-15
Rebase x86

nicoTolly:
Move x86 sources into its own backend crate


src/codegen/variable.rs
```diff
@@ -207,58 +206,4 @@ mod tests {
     {
         content.iter().cloned().collect()
     }
-
-    /// Ensures chained aliases work correctly.
-    #[test]
```
ulysseB: Why is this test removed ?

src/device/mod.rs
```diff
@@ -1,6 +1,7 @@
 //! Code generation and candidate evaluation for specific targets.
 pub mod fake;
-pub mod x86;
+#[cfg(feature = "mppa")]
+pub mod mppa;
```
ulysseB: Why are you adding back mppa as a module here ?

src/codegen/variable.rs
nicoTolly: It is the circular dependency we talked about

src/device/mod.rs
nicoTolly: Merge error

src/codegen/variable.rs
ulysseB: Ok. Fine for me.

nicoTolly:
bors retry

nicoTolly:
bors retry

nicoTolly:
bors retry

nicoTolly:
bors retry

nicoTolly:
bors retry

PR #218, opened 2019-03-20
Minimal support for Maxwell and Pascal architectures

Elarnon:
This patch adds minimal support for Maxwell and Pascal CUDA GPUs, by
adding information from the CUDA C Programming Guide on those GPUs and
overall ensuring that generated kernels can run on them.  The
performance model is not yet updated for those architectures, and as
such may give (wildly) incorrect lower bounds -- which shouldn't prevent
the generated kernels from being correct still.

In addition, this also fixes a pre-existing bug where constraints on inner block dimensions were not properly computed due to a division rounding error, which caused crashes when running Telamon on Maxwell/Pascal gpus.

### DIFF
--- a/backend/cuda/src/characterize/gpu.rs
+++ b/backend/cuda/src/characterize/gpu.rs
@@ -1,4 +1,28 @@
 //! Builds the description of a GPU.
+//! Unless otherwise specified, most of the information comes from the "CUDA C Programming Guide",
+//! available online at https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html
+//! For the sake of brevity, I will refer to this document as "the CPG".
+//!
+//! Some additional information comes from various nvidia whitepapers:
+//!
+//!  - For Kepler:
+//!    https://la.nvidia.com/content/PDF/product-specifications/GeForce_GTX_680_Whitepaper_FINAL.pdf
+//!
+//!  - For Maxwell:
+//!    https://international.download.nvidia.com/geforce-com/international/pdfs/GeForce-GTX-750-Ti-Whitepaper.pdf
+//!
+//!  - For Pascal:
+//!    https://images.nvidia.com/content/pdf/tesla/whitepaper/pascal-architecture-whitepaper.pdf
+//!
+//!  - For Volta:
+//!    https://images.nvidia.com/content/volta-architecture/pdf/volta-architecture-whitepaper.pdf
+//!
+//!  - For Turing:
+//!    https://www.nvidia.com/content/dam/en-zz/Solutions/design-visualization/technologies/turing-architecture/NVIDIA-Turing-Architecture-Whitepaper.pdf
+//!
+//! Note: Sometimes compute capabilities are used, and sometimes architecture names are used.  The
+//! mapping between those is described in section 2.5 "Compute Capability" of the CPG.
+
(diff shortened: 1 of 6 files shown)


backend/cuda/src/characterize/gpu.rs
```diff
@@ -73,56 +97,87 @@ pub fn functional_desc(executor: &Executor) -> Gpu {
...
+        (3, 0) | (3, 2) | (3, 5) => false,
+        // Some 3.5 devices allow L1 as well.  Ignore them.
+        (3, 7) => true,
+        (5, 0) => false,
+        (5, 2) | (5, 3) => true,
+        // FIXME: nvidia docs are not clear on this for compute capabilities 6.x and 7.x, saying
+        // "Global memory behaves the same way as devices of compute capability 5.x".  But CC 5.0
+        // behaves differently from CC 5.2+ and does not allow using L1 for global memory -- I will
+        // assume this means to use the CC 5.2+ behavior and allow L1 cache, but we should check
+        // this.
```
ulysseB: Yes they do. This corresponds to the `ca` flag, which is supported on pascal https://docs.nvidia.com/cuda/pascal-tuning-guide/index.html#l1-cache

backend/cuda/src/gpu.rs
```diff
@@ -476,13 +476,12 @@ impl device::Device for Gpu {
             | ir::Operator::TmpSt(..) => InstFlag::COHERENT,
             _ => panic!("invalid memory access operator"),
         };
-        // Remove the `CACHE_READ_ONLY` option if the gpu does not support `ld.nc`.
-        if !self.allow_nc_load {
+        // TODO(model): CACHE_READ_ONLY and CACHE_SHARED are currently not supported by the
+        // performance model.  Disable them, even if the hardware supports them.
+        if !self.allow_nc_load || true {
```
ulysseB: Be warned, this might have a good impact on performance.But I don't have a better solution for now.

Elarnon: If I don't put this here, we crash on the `assert!` in `load_desc` and `store_desc` anyways :)  The goal of this PR is to get Telamon running on those hardware and leave the "hard" problems for later.

ulysseB: Yes I know. I just wanted you to be aware of that.

Elarnon: Ah right -- I thought you were saying "if we don't put this here we will be faster".  Noted that performance may be strongly impacted; let's chat offline about what would be needed to include those accesses in the model sometime.

backend/cuda/src/characterize/gpu.rs
```diff
@@ -73,56 +97,87 @@ pub fn functional_desc(executor: &Executor) -> Gpu {
 }
 
 /// Returns true if the GPU allows non-coherent loads.
+/// Not 100% sure what this referred to innitially, but we don't use it anyways.  I take this to
+/// correspond to the availability of the read-only cache which also backs up the texture cache.
```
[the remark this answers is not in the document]
Elarnon: It does look like it should be the same, although I a am not sure if/how the unified cache is segregated.  I will also update the comment to note this.

Elarnon:
bors r=ulysseB

Elarnon:
bors retry

PR #219, opened 2019-03-21
Add ability to dump and load kernel implementations

nicoTolly:
Added a binary in kernels that allows to generate a dump of an implementation and a function that takes a dump, execute it and check the results. This allows more flexibility in testing new backends - reruns of candidate that failed before.


kernels/Cargo.toml
```diff
@@ -14,22 +15,26 @@ num = "0.2.0"
...
-telamon-x86 = { path = "../backend/x86" }
 
 [features]
 cuda = [
     "telamon-cuda",
     "cuda-sys",
 ]
 
+[[bin]]
+name = "kernel_dump"
```
Elarnon: This is should no longer be needed with 2018 edition.

kernels/src/kernel.rs
```diff
@@ -56,6 +57,95 @@ pub trait Kernel<'a>: Sized {
...
+                path = format!(
+                    "kernel_dump/{}_{}_{}.log",
+                    Self::name(),
+                    ctx.device().name(),
+                    increment
+                );
+            } else {
+                let mut file = File::create(path).unwrap();
+                file.write_all(&dump).unwrap();
+                break;
```
Elarnon: You probably should `flush()` the file here to be safe.

kernels/src/bin/kernel_dump.rs
```diff
@@ -0,0 +1,17 @@
+use telamon_kernels::{linalg, Kernel};
```
Elarnon: Some documentation on this file that it is used to generate kernel dumps to the test directory (and usually should not be run) would be useful.

kernels/src/kernel.rs
```diff
@@ -56,6 +57,95 @@ pub trait Kernel<'a>: Sized {
...
+                    .unwrap();
+                candidate = leaf;
+                break;
+            }
+        }
+        let dump = bincode::serialize(&candidate.actions).unwrap();
+        // Create dir kernel_dump if it does not exists
+        std::fs::create_dir_all("kernel_dump").unwrap();
+        let mut path =
+            format!("kernel_dump/{}_{}.log", Self::name(), ctx.device().name());
```
Elarnon: There must be some de-sync between commits here because that is not the path of the files that are included in this commit, nor the files used by the tests?

BTW it would make sense for `generate_dump` to take the path (or even a `Write` instance) where the log should be dumped, and have the file-related logic in the `kernel_dump` executable instead.  Would make it easier to reuse for other purposes (eg dumping best candidate at end of exploration).

kernels/src/kernel.rs
```diff
@@ -56,6 +57,95 @@ pub trait Kernel<'a>: Sized {
...
+            }
+        }
+    }
+
+    /// Takes a path to a log and execute it. Caller is responsible for making sure that the log
+    /// corresponds to the kernel and context being executed
+    fn execute_log<AM, P: AsRef<Path>>(
+        params: Self::Parameters,
+        ctx: &mut AM,
+        dump_path: P,
```
Elarnon: Taking a `Read` instance here would be a bit more flexible but I don't think it'd be needed anyways.

Elarnon: However what *would* be more flexible (and useful) would be to have a `load_candidate` function returning the loaded candidate from the dump instead, and have the evaluation-related code in the test file.  It'd be useful for also benchmarking a loaded kernel, not just checking correctness (I actually need this, so I can make a follow-up PR on that).

kernels/src/kernel.rs
nicoTolly: Yes, I should have make that clearer. The intent was that dumps were generate in a kernel_dump directory without erasing the previous ones, and that the user is responsible for choosing one specific dump and to place it in the log_to_test directory. I was not sure about this design anyway.

Elarnon: I think it's better to generate them in place (skipping with a warning if there is already an existing file -- if the user wants a new one they should first `mv` the old one).  This avoids the extra mental overhead of having to move the file(s), which we will probably have forgotten by the time we next need to run this -- probably this will only be run when we want to generate new kernels so it makes sense that just running `kernel_dump` would add the new kernel logs and do nothing else.

Elarnon:
(Also, please change the title of the PR -- which is used as commit shortname by bors -- to something more explicit, e.g. "Add ability to dump and load kernel implementations")

kernels/src/bin/kernel_dump.rs
```diff
@@ -0,0 +1,30 @@
...
+    ($kernel:ty, $num_tests:expr, $params:expr) => {
+        let _ = env_logger::try_init();
+        let mut context = x86::Context::default();
+        let mut path = format!("kernel_dump/x86/{}.dump", <$kernel>::name());
+        let mut increment = 0;
+        loop {
+            if Path::new(&path).exists() {
+                increment += 1;
+                path =
+                    format!("kernel_dump/x86/{}_{}.dump", <$kernel>::name(), increment);
```
Elarnon: What does adding a suffix here gives us wrt just skipping the generation?  I don't see a scenario where this is helpful: if I want the new dump I need to move something in both cases; if I don't want the new dump I have nothing to do if skipping but I need to remove it with the suffix.

nicoTolly: Ok, let's go for skipping - I'm also adding a warn!("Skipping generation as a dump already exists")

nicoTolly:
bors r=elarnon r=ulysseB

nicoTolly:
bors retry

nicoTolly:
bors retry

PR #222, opened 2019-03-26
Embedding Namer in NameMap entirely

nicoTolly:
We used to have a RefCell<&mut Namer> inside of NameMap struct, but we use Namer both directly and through NameMap in printer. This leads to unconveniance like putting use of NameMap inside a block and is troublesome when calls to Namer can be a bit more complicated as it will be in Mppa (call to Namer for both wrapper and kernel code). NameMap now takes directly a mutable reference to a generic N implementing Namer (we very often use a &mut NameMap anyway, so it does not make much difference) and can provide immutable or mutable reference to this N.
Printer of cuda and x86 has been modified accordingly.

### DIFF
--- a/src/codegen/name_map.rs
+++ b/src/codegen/name_map.rs
@@ -152,11 +152,11 @@ impl<'a, 'b, N: Namer> NameMap<'a, 'b, N> {
         name_map
     }
 
-    pub fn get_namer(&self) -> &N {
+    pub fn namer(&self) -> &N {
         &*self.namer
     }
 
-    pub fn get_mut_namer(&mut self) -> &mut N {
+    pub fn namer_mut(&mut self) -> &mut N {
         self.namer
     }
 
--- a/src/codegen/name_map.rs
+++ b/src/codegen/name_map.rs
@@ -33,9 +33,9 @@ pub trait Namer {
 
 /// Maps variables to names.
 // TODO(cc_perf): use an arena rather that ref-counted strings
-pub struct NameMap<'a, 'b> {
+pub struct NameMap<'a, 'b, N: Namer> {
     /// Provides fresh names.
-    namer: std::cell::RefCell<&'b mut Namer>,
+    namer: &'b mut N,
     /// Keeps track of the names of the variables used in the kernel
     variables: FnvHashMap<ir::VarId, VariableNames>,
     /// Keeps track of the name of the values produced by instructions.
@@ -62,9 +62,9 @@ pub struct NameMap<'a, 'b> {
     side_effect_guard: Option<RcStr>,
 }
 
-impl<'a, 'b> NameMap<'a, 'b> {
+impl<'a, 'b, N: Namer> NameMap<'a, 'b, N> {
     /// Creates a new `NameMap`.
-    pub fn new(function: &'a Function<'a>, namer: &'b mut Namer) -> Self {
+    pub fn new(function: &'a Function<'a>, namer: &'b mut N) -> Self {
         let mut mem_blocks = FnvHashMap::default();
         // Setup parameters names.
         let params = function
@@ -106,7 +106,7 @@ impl<'a, 'b> NameMap<'a, 'b> {
         }
         let variables = VariableNames::create(function, namer);
         let mut name_map = NameMap {
-            namer: std::cell::RefCell::new(namer),
+            namer,
             insts: FnvHashMap::default(),
             variables,
             num_loop: 0,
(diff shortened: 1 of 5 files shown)


src/codegen/name_map.rs
```diff
@@ -152,15 +152,23 @@ impl<'a, 'b> NameMap<'a, 'b> {
         name_map
     }
 
+    pub fn get_namer(&self) -> &N {
```
Elarnon: This would typically be called either `namer()` (if we consider this is a field of the `NameMap`, as per [the naming conventions](https://doc.rust-lang.org/1.0.0/style/style/naming/README.html#getter/setter-methods-[rfc-344])) or `get_ref` (if we consider the `NameMap` to be "wrapping" the `Namer`, consistent with e.g. [`BufWriter`](https://doc.rust-lang.org/std/io/struct.BufWriter.html)).

src/codegen/name_map.rs
```diff
@@ -152,15 +152,23 @@ impl<'a, 'b> NameMap<'a, 'b> {
...
+        &*self.namer
+    }
+
+    pub fn get_mut_namer(&mut self) -> &mut N {
```
Elarnon: Same as above, either `namer_mut` or `get_mut`

src/codegen/printer.rs
```diff
@@ -305,15 +310,19 @@ pub trait Printer {
...
-            let base_components = level.base.components().map(|v| namer.name(v));
-            let base = match base_components.collect_vec()[..] {
-                [ref base] => base.to_string(),
+            let base_components = level
+                .base
+                .components()
+                .map(|v| namer.name(v).into_owned())
+                .collect_vec();
+            let base = match base_components[..] {
+                [ref base] => base.clone(),
```
Elarnon: Since you talked about performance on Slack:
 - In the previous code there is a single unconditional clone (in `.to_string()`) for a base component and no clone at all for lhs+rhs components
 - In the new code there is one unconditional clone (in `.clone()` called on the `String`) and one potential clone (in `.into_owned()`) for base components, and two potential clones (in `.into_owned()`) for lhs+rhs components

All of this is probably premature optimisation anyways, but as far as I can tell -- there is actually no need to use `String` here and everything could be done using `Cow`.

src/codegen/printer.rs
```diff
@@ -347,20 +356,22 @@ pub trait Printer {
...
         if fun.block_dims().is_empty() {
             return;
         }
-        let addr = namer.name_addr(block.id());
+        let addr = namer.name_addr(block.id()).into_owned();
         let addr_type = Self::lower_type(Type::PtrTo(block.id()), fun);
-        let size = namer.name_size(block.local_size(), Type::I(32));
+        let size = namer
+            .name_size(block.local_size(), Type::I(32))
+            .into_owned();
```
Elarnon: Why do we need those extra `.into_owned()`?

src/codegen/name_map.rs
```diff
@@ -152,15 +152,23 @@ impl<'a, 'b> NameMap<'a, 'b> {
         name_map
     }
 
+    pub fn namer(&self) -> &N {
+        &*self.namer
```
Elarnon: BTW this is serious nitpicking -- but I believe `self.namer` would be enough.

src/codegen/printer.rs
Elarnon: Ah -- I guess this is because now `gen_XXX` require a mutable reference on the `NameMap`...

src/codegen/printer.rs
Elarnon: Actually there probably *is* now, because of `gen_name` taking a mutable reference now...

nicoTolly: This is the problem I ran into that made me ask the question about into_owned in the first place. It seems that without into_owned we somehow keep a reference on namer that then forbids the call of namer.gen_name which takes a mutable ref on namer. This problem is solved by into_owned in a way that I don't really understand - I failed to isolate the problem. I will try to find another way

Elarnon: As said below I don't think it matters -- as far as I know this is not performance-critical code (if it is we need to do something).  `Cow` can keep a reference into the name map, so it needs to be converted into a string (with `into_owned`) before using the name map mutably (Rust has no way of knowing that *actually* the reference that the `Cow` holds will never overlap with what the `gen_XXX` functions will modify).

nicoTolly:
bors r=elarnon

nicoTolly:
bors retry

nicoTolly:
bors retry

PR #223, opened 2019-03-28
Dump Parameters along with decisions

nicoTolly:
Until now, the user was responsible to provide the right parameters along with a dump in which decisions were serialized. Now we dump both parameters and decisions, which allows the user to care only for using the right kernel with the right dump. A binary allowing to execute a dump by reading command line argument was also added. Not having to provide Parameters comes in handy here, as the parameters have different types depending on the kernel, which made the dynamic dispatch difficult.


kernels/src/bin/exec_dump.rs
```diff
@@ -0,0 +1,29 @@
...
+    assert!(args.len() == 3);
+    let kernel_name = &args[1];
+    let binary_path = &args[2];
+
+
+    let _ = env_logger::try_init();
+    let mut context = x86::Context::default();
+
+    let mut file = std::fs::File::open(binary_path).expect("Invalid dump path");
+    dispatch_kernel!(kernel_name as &str, execute_dump, &mut context, &mut file);
```
Elarnon: What's the reasoning behind using a macro here, since it is only called once?  Inlining it would make the code more readable and explicit IMO.

nicoTolly:
bors r=elarnon

nicoTolly:
bors retry

nicoTolly:
bors retry

PR #224, opened 2019-04-01
Add a new Mppa backend

nicoTolly:
This PR adds an MPPA backend. The code generation is quite similar to the x86 one, but here telajax is used as "runtime". This is still buggy though : kernels tend to segfault, to hang or to segfault when we run a lot of them. We probably don't want to rely on telajax in the long run anyway.


backend/mppa/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+version = "1.0.0"
+edition = "2018"
+
+[build-dependencies]
+cc = "1.0"
+
+[dependencies]
+crossbeam = "0.7"
+lazy_static = { version = "1.0" }
+errno = { version = "0.2", optional = true }
```
Elarnon: Pretty sure if you put this as `optional` here then don't have any feature that depend on it, it will never be used?

backend/x86/src/printer.rs
```diff
@@ -31,20 +31,17 @@ impl X86printer {
...
-                s.push_str(
-                    &(0..n)
-                        .map(|i| format!("{}{}", prefix, i))
-                        .collect_vec()
-                        .join(", "),
-                );
-                s.push_str(";\n  ");
-                s
-            }
+        let print_decl = |(&t, &n)| {
```
Elarnon: Why do we no longer ignore pointers (or rather why did we ignore them previously)?

backend/mppa/src/template/host.c.template
```diff
@@ -0,0 +1,91 @@
...
+  return NULL;
+}}
+
+void entry_point({cl_arg_def}, void* __timer_ptr) {{
+  // KERNEL {id}
+  unsigned long timestamp;
+  void* args[{n_arg}];
+  {build_ptr_struct};
+  pthread_barrier_t barrier;
+  printf("KERNEL {id} START\n");
```
Elarnon: Should we keep this (looks like a debug print to me, but maybe not)?

backend/mppa/src/mppa.rs
```diff
@@ -0,0 +1,142 @@
...
+pub struct Mppa;
+
+impl device::Device for Mppa {
+    fn print(&self, _fun: &Function, out: &mut Write) {
+        unwrap!(write!(out, "Basic MPPA"));
+    }
+
+    fn check_type(&self, t: Type) -> Result<(), ir::TypeError> {
+        match t {
+            Type::I(i) | Type::F(i) if i == 32 || i == 64 => Ok(()),
```
Elarnon: The namer supports more types than these; is it expected?

backend/mppa/src/lib.rs
```diff
@@ -0,0 +1,98 @@
...
+                *entry += 1;
+                name
+            }
+        }
+    }
+
+    fn name_float(&self, val: &Ratio<BigInt>, len: u16) -> String {
+        assert!(len <= 64);
+        let f = unwrap!(val.numer().to_f64()) / unwrap!(val.denom().to_f64());
+        format!("{:.5e}", f)
```
Elarnon: Can't we lose precision here?

backend/mppa/src/context.rs
```diff
@@ -0,0 +1,341 @@
...
+use telamon::codegen::{Function, NameMap, ParamVal};
+use telamon::device::Context as ContextTrait;
+use telamon::device::{self, ArrayArgument, AsyncCallback, EvalMode, ScalarArgument};
+use telamon::explorer;
+use telamon::ir;
+use utils::unwrap;
+use utils::*;
+
+lazy_static! {
+    static ref ATOMIC_USIZE: AtomicUsize = AtomicUsize::new(0);
```
Elarnon: We probably can figure out a better name for this.  Also I don't think you need `lazy_static` here -- `const A_BETTER_NAME: AtomicUsize = std::atomic::ATOMIC_USIZE_INIT;`.

(In fact I believe `AtomicUsize::new` is now a `const fn` and so `const A_BETTER_NAME: AtomicUsize = AtomicUsize::new(0);` should work as well)

backend/mppa/src/context.rs
```diff
@@ -0,0 +1,341 @@
...
+        device::ScalarArgument::raw_ptr(&**self as &dyn ScalarArgument)
+    }
+
+    fn as_size(&self) -> Option<u32> {
+        device::ScalarArgument::as_size(&**self as &dyn ScalarArgument)
+    }
+}
+
+/// Wrapper around Buffer
+/// We need it to implement Argument for Buffer (orphan rule)
```
Elarnon: I believe it's actually to implemente `ArrayArgument` for Buffer?  `impl Argument for Buffer` should work since `Argument` is defined in the current crate.

backend/mppa/src/context.rs
```diff
@@ -0,0 +1,341 @@
...
+    GlobalMem(MppaArray),
+    Size(u32),
+    External(*const libc::c_void),
+}
+
+impl KernelArg {
+    fn raw_ptr(&self) -> *const libc::c_void {
+        match self {
+            KernelArg::GlobalMem(mem) => mem.raw_ptr(),
+            KernelArg::Size(size) => size as *const _ as *const libc::c_void,
```
Elarnon: Why the double cast? `size as *const libc::c_void` doesn't work?

Elarnon: Ah no because `size` is an `&u32`.  Please use `size as *const u32 as *const libc::c_void` in that case, `*const _` is a very dangerous type.

backend/mppa/src/context.rs
```diff
@@ -0,0 +1,341 @@
...
+}
+
+pub struct Context {
+    device: mppa::Mppa,
+    executor: &'static telajax::Device,
+    parameters: FnvHashMap<String, Arc<Argument>>,
+    writeback_slots: ArrayQueue<MppaArray>,
+}
+
+unsafe impl Sync for Context {}
```
Elarnon: What is preventing the auto deriving of `Sync` for `Context`?

backend/mppa/src/context.rs
```diff
@@ -0,0 +1,341 @@
...
+        self.executor.execute_kernel(&mut kernel).unwrap();
+        let out_mem = if let KernelArg::GlobalMem(mem) = kernel_args.pop().unwrap() {
+            mem
+        } else {
+            panic!()
+        };
+        // FIXME: We get obviously wrong timings, fix that
+        let ptr_i8 = out_mem.read_i8().as_ptr();
+        let res: usize =
+            unsafe { *std::mem::transmute::<*const i8, *const usize>(ptr_i8) };
```
Elarnon: Why `transmute` here?  How do we know the resulting value contains an `usize`?  At the very least there should be an `assert!` that the `Vec<i8>` read from `out_mem.read_i8()` has length equal to `std::mem::size_of::<usize>()` -- but probably using `u64::from_ne_bytes` or equivalent would be better.

Elarnon: (also there is no need for transmute afaik -- `ptr_i8 as *const usize` should work as well -- but the remark about asserting the length of the vector still stands)

backend/mppa/src/context.rs
```diff
@@ -0,0 +1,341 @@
...
+                    .spawn(move |_| inner(&mut evaluator)));
+            }
+            // Start the evaluation thread.
+            let eval_thread_name = "Telamon - CPU Evaluation Thread".to_string();
+            unwrap!(scope.builder().name(eval_thread_name).spawn(move |_| {
+                while let Ok((candidate, mut kernel, callback)) = recv.recv() {
+                    // TODO: measure time directly on MPPA
+                    let t0 = Instant::now();
+                    self.executor.execute_kernel(&mut kernel).unwrap();
+                    let t = Instant::now() - t0;
```
Elarnon: `let d = t0.elapsed();`

backend/mppa/src/context.rs
```diff
@@ -0,0 +1,341 @@
...
+    'a: 'b,
+{
+    context: &'b Context,
+    sender: mpsc::SyncSender<AsyncPayload<'a, 'b>>,
+}
+
+impl<'a, 'b, 'c> device::AsyncEvaluator<'a, 'c> for AsyncEvaluator<'a, 'b>
+where
+    'a: 'b,
+    'c: 'b,
```
Elarnon: *cries in lifetimes*

backend/mppa/src/context.rs
nicoTolly: Something is bothering me a little here. from_ne_bytes takes an array and not a slice, meaning that we need a copy in a way or another. It doesn't matter with respect to performance (it's a small copy anyway) but I feel like it adds a bit of rather ugly code (either reading vector elements or doing an unsafe memcpy). Transmute is unsafe but it isn't troubling me much as it conveys what I want to do here : reinterpret these bytes as an u64. Rust has to trust me here anyway so I feel like it is inherently unsafe. We already have a guarantee on length as slot is a Buffer<i8> of length 8 (see declaration in Context::new()) so we read 8 bytes whatever we do here. Long story short, I feel like transmute conveys better semantic here.

nicoTolly: However, we may have a problem of dangling pointer here

nicoTolly: We do

Elarnon: Yes there is some gymnastics needed for `from_ne_bytes` but it would allow safe code here :)  If you think `from_ne_byte` makes things more confusing and `unsafe` is better (which it may be, provided the proper checks are in place -- which doesn't look like they are if there is a dangling pointer, although I am not sure I understand which pointer is dangling), then `unsafe { *(ptr_i8 as *const usize) }` is the solution, not `transmute`.

Also you want to reinterpret as `u64` you should use `u64`, not `usize`.

nicoTolly: out_mem.read_i8().as_ptr() creates a dangling pointer if I'm not mistaken

Elarnon: Ah, right, it indeed does!

backend/mppa/src/lib.rs
```diff
@@ -77,7 +77,7 @@ impl codegen::Namer for Namer {
     fn name_float(&self, val: &Ratio<BigInt>, len: u16) -> String {
         assert!(len <= 64);
         let f = unwrap!(val.numer().to_f64()) / unwrap!(val.denom().to_f64());
-        format!("{:.5e}", f)
+        format!("{}", f.to_string())
```
Elarnon: You can just do: `f.to_string()` (or `format!("{}", f)` which is equivalent.

nicoTolly:
bors r=elarnon

nicoTolly:
bors retry

nicoTolly:
bors retry

nicoTolly:
bors retry

PR #228, opened 2019-04-01
Split the explorer in a separate crate

Elarnon:
This patch extracts the `telamon::explorer` module into a separate
`telamon_explorer` crate, responsible for everything related to the
search algorithms.

This should help with compilation times since modifying the explorer
will no longer require recompiling the rest of the `telamon` crate (and
in particular not the generated `choices.rs`).


Elarnon:
Marking as draft until #226 and #227 are reviewed; then I will retarget this PR towards master, but this is mostly done already -- comments are accepted :)

PR #229, opened 2019-04-01
Remove lifetime in Candidate

Elarnon:
This patch removes the lifetime in `ir::Function` due to holding
references to the signature, parameters, and device.  The references are
replaced with atomically reference-counted pointers instead.  Since
those internal structures are rarely copied (they are already behind
other `Arc`s which get themselves cloned, except when during lowering),
there should be virtually no performance difference, but it makes
various APIs nicer -- and limits the lifetime hell we would occasionally
find ourselves in.

The PR touches a large amount of files, but most of the changes in there
are fairly straightforward -- it's just removing lifetimes and
(sometimes) adding `Arc`s.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,7 +50,7 @@ linked-list = "0.0.3"
 log = "0.4.1"
 matches = "0.1.6"
 ndarray = "0.11.2"
-num = "0.2.0"
+num = { version = "0.2.0", features = ["serde"] }
 num_cpus = "1.8.0"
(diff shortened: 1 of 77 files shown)


Elarnon:
bors retry

Elarnon:
bors retry

Elarnon:
bors r=nicoTolly

Elarnon:
bors retry

Elarnon:
bors retry

PR #230, opened 2019-04-02
Add conversion tool for Aftermath trace files

andidr:
Add a conversion tool creating a trace file for the [Aftermath performance analysis tool](https://www.aftermath-tracing.com) from a Telamon log file

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,6 +40,7 @@ path = "telamon-gen"
 [dependencies]
 binary-heap-plus = "0.1.4"
 boxfnonce = "0.1.0"
+byteorder = "1"
 config = "0.8.0"
 crossbeam = "0.3.2"
 env_logger = "0.5.9"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,12 +8,6 @@ edition = "2018"
 readme = "README.md"
 repository = "https://github.com/ulysseB/telamon/"
 
-[[bin]]
-name = "expandconfig"
-
-[[bin]]
-name = "parse_event_log"
-
 [[bench]]
 name = "descent"
 harness = false
--- a/src/bin/aftermath_convert_log.rs
+++ b/src/bin/aftermath_convert_log.rs
@@ -1,0 +1,278 @@
+///! Aftermath_convert_log converts a Telamon logfile into a trace
+///! file that can be loaded into the Aftermath trace analysis tool
+///! from https://www.aftermath-tracing.com
+use log::debug;
+use std::collections::HashSet;
+use std::io;
+use std::path::PathBuf;
+use std::time::Duration;
+use structopt::StructOpt;
+use telamon::explorer::eventlog::EventLog;
+use telamon::explorer::mcts::{Event, Message, Policy};
+use telamon::offline_analysis::aftermath::TraceWriter;
+use telamon::offline_analysis::tree::CandidateTree;
+
+#[derive(Debug, StructOpt)]
+#[structopt(name = "aftermath_convert_log")]
+struct Opt {
+    #[structopt(
+        parse(from_os_str),
+        short = "i",
+        long = "input",
+        default_value = "eventlog.tfrecord.gz"
+    )]
+    eventlog: PathBuf,
+
+    #[structopt(
+        parse(from_os_str),
+        short = "o",
+        long = "output",
+        default_value = "eventlog.ost"
+    )]
+    output: PathBuf,
+}
+
+fn main() -> io::Result<()> {
+    env_logger::try_init().unwrap();
+
+    let opt = Opt::from_args();
+    let mut tw = TraceWriter::new(&opt.output.as_path()).unwrap();
+    let mut thread_ids = HashSet::new();
+    let mut t = CandidateTree::new();
+
+    // Write file header and declare all Aftermath frame types used in
+    // the trace
+    tw.write_default_header()?;
+    tw.write_default_frame_ids()?;
+
+    // Process log messages
+    for record_bytes in EventLog::open(&opt.eventlog)?.records() {
+        match bincode::deserialize(&record_bytes?).unwrap() {
+            // Discovery of a new node
+            Message::Node {
+                id,
+                parent,
+                bound,
+                children,
+                discovery_time,
+            } => {
+                debug!("Node (ID {}) [discovery time: {:?}]", id, discovery_time);
+
+                t.extend(
+                    id.into(),
+                    discovery_time.clone(),
+                    parent,
+                    bound.clone(),
+                    &mut children.clone(),
+                );
+            }
+
+            // Series of timed events performed by one thread
+            Message::Trace { thread, events } => {
+                // ID of the node that was last selected by an action;
+                // The root node is implicitly selected at the start
+                // of a thread trace
+                let mut curr_node_id = t.get_root().id();
+                let thread_id = thread
+                    .trim_start_matches("ThreadId")
+                    .trim_matches(|p| p == '(' || p == ')')
+                    .parse::<u32>()
+                    .unwrap();
+                let mut trace_start: Option<Duration> = None;
+                let mut trace_end: Option<Duration> = None;
+
+                // If the thread is yet unknown, start a new Aftermath
+                // event collection for its events
+                if thread_ids.insert(thread_id) {
+                    tw.write_event_collection(thread_id)?;
+                }
+
+                debug!("Trace (thread {}):", thread_id);
+                debug!("  Implicit SelectNode {}", curr_node_id);
+
+                for timed_event in &events {
+                    let start = timed_event.start_time;
+                    let end = timed_event.end_time;
+
+                    // Set start timestamp of the thread trace to the
+                    // beginning of the first event of the trace and
+                    // the end timestamp to the end of the last event
+                    if trace_start.is_none() {
+                        trace_start = Some(start.clone());
+                    }
+
+                    trace_end = Some(end.clone());
+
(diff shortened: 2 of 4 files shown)


Cargo.toml
```diff
@@ -14,6 +14,9 @@ name = "expandconfig"
 [[bin]]
 name = "parse_event_log"
 
+[[bin]]
+name = "aftermath_convert_log"
```
Elarnon: Not needed anymore in rust 2018, remove this (and remove the other `[[bin]]` sections while you are at it).

src/offline_analysis/aftermath.rs
```diff
@@ -0,0 +1,619 @@
...
+const AM_TELAMON_CANDIDATE_FLAG_DEADEND: u32 = (1 << 3);
+const AM_TELAMON_CANDIDATE_FLAG_PERFMODEL_BOUND_VALID: u32 = (1 << 4);
+const AM_TELAMON_CANDIDATE_FLAG_SCORE_VALID: u32 = (1 << 5);
+
+const AM_TELAMON_CANDIDATE_EVALUATION_FLAG_SCORE_VALID: u8 = (1 << 0);
+
+const HIERARCHY_DEFAULT_ID: u32 = 0;
+const HIERARCHY_ROOT_NODE_ID: u32 = 1;
+
+enum CandidateKillActionCause {
```
Elarnon: Deriving `Debug`, `Copy` and `Clone` almost always make sense (here and elsewhere).

src/offline_analysis/aftermath.rs
```diff
@@ -0,0 +1,619 @@
...
+
+impl From<FrameType> for u32 {
+    fn from(v: FrameType) -> Self {
+        v as u32
+    }
+}
+
+/// Converts a Duration to a u64 in nanoseconds
+fn duration_ns(d: Duration) -> u64 {
+    d.as_secs() * 1000000000u64 + d.subsec_nanos() as u64
```
Elarnon: Use `1_000_000_000u64` for readability

src/offline_analysis/aftermath.rs
```diff
@@ -0,0 +1,619 @@
...
+}
+
+/// Converts a Duration to a u64 in nanoseconds, or to 0 if the Option
+/// is None
+fn duration_ns_default(d: Option<Duration>) -> u64 {
+    d.map_or(0, |d| duration_ns(d))
```
Elarnon: `.map_or(0, duration_ns)` is only three characters more than `duration_ns_default` -- probably makes sense to write it directly.

src/offline_analysis/aftermath.rs
```diff
@@ -0,0 +1,619 @@
...
+}
+
+/// Writer producing trace files in Aftermath format
+pub struct TraceWriter {
+    bw: BufWriter<File>,
```
Elarnon: Doesn't really matter here, but FYI -- such structures are usually parameterize with an underlying writer (ie `TraceWriter<W> { writer: W }`) then `from_writer` constructor is provided, possibly with helpers e.g. `from_path`.  This lets the user select how they want to use the writer (eg maybe write into stdout or a pipe).

src/offline_analysis/aftermath.rs
```diff
@@ -0,0 +1,619 @@
...
+    /// truncated.
+    pub fn new(path: &Path) -> Result<Self, io::Error> {
+        let fp = File::create(path)?;
+        Ok(TraceWriter {
+            bw: BufWriter::new(fp),
+        })
+    }
+
+    /// Writes a u8 to the output file
+    fn write_u8(&mut self, val: u8) -> Result<(), io::Error> {
```
Elarnon: I'm not 100% convinced we need the wrapper functions around the `.bw` functions (i.e. `self.write_u32(x)` vs `self.bw.write_u32::<LE>(x)` but OK.  I guess this encapsulates the internal aftermath endianness.

BTW you don't need to bother with self-explanatory comments (e.g. "Writes a u8" for a method named `write_u8`) for private functions (doing it is fine, but not doing it also is).

src/bin/aftermath_convert_log.rs
```diff
@@ -0,0 +1,275 @@
...
+    )]
+    output: PathBuf,
+}
+
+fn main() -> io::Result<()> {
+    env_logger::try_init().unwrap();
+
+    let opt = Opt::from_args();
+    let mut tw = TraceWriter::new(&opt.output.as_path()).unwrap();
+    let mut thread_ids = BitSet::new();
```
Elarnon: Isn't using a bit set here instead of a regular set premature optimization?  I'd rather not have the external dependency to be fair.

src/bin/aftermath_convert_log.rs
```diff
@@ -0,0 +1,275 @@
...
+                );
+            }
+
+            // Series of timed events performed by one thread
+            Message::Trace { thread, events } => {
+                // ID of the node that was last selected by an action;
+                // The root node is implicitly selected at the start
+                // of a thread trace
+                let mut curr_node_id = t.get_root().id();
+                let thread_id = scan_fmt!(&thread, "ThreadId({})", usize).unwrap();
```
Elarnon: Not sure it's worth it to use a whole external crate to avoid writing

```rust
let thread_id = thread.trim_start_matches("ThreadId").trim_matches(|p| p == '(' || p == ')').parse::<usize>().unwrap();
```

src/offline_analysis/aftermath.rs
```diff
@@ -0,0 +1,619 @@
+///! Functions converting recreated tree and events into Aftermath
+///! trace
+use crate::explorer::mcts::{CauseOfDeath, EdgeIndex, NodeId, Selector};
+use crate::offline_analysis::tree::{CandidateNode, CandidateTree};
+use byteorder::{LittleEndian, WriteBytesExt};
+use std::fs::File;
+use std::io;
+use std::io::{BufWriter, Write};
```
Elarnon: Usually we do: `use std::io::{self, BufWriter, Write}` instead of the two lines above.

src/offline_analysis/aftermath.rs
```diff
@@ -0,0 +1,619 @@
...
+        &mut self,
+        thread_id: u32,
+        candidate_id: NodeId,
+        start: Duration,
+        end: Duration,
+        cause: &CauseOfDeath,
+    ) -> Result<(), io::Error> {
+        let cause_numeric: u8;
+
+        match cause {
```
Elarnon: In Rust there is no distinction between expressions and statements so you can write (idiomatically):

```rust
let cause = match cause {
    CauseOfDeath::Constraints => CandidateKillActionCause::Constraints.into(),
    CauseOfDeath::Constraints => CandidateKillActionCause::PerfModel.into(),
    CauseOfDeath::Constraints => CandidateKillActionCause::Backtrack.into(),
};
```

Elarnon:
Oh, and please update the title of the PR -- it ends up as the title for the merge commit, and "Andi/aftermath" is not very evocative :)

src/offline_analysis/aftermath.rs
andidr: CandidateKillActionCause is private to the module, only used once and will never be used in any other way, so I see no point in deriving from Debug, Copy and Clone.

src/offline_analysis/aftermath.rs
andidr: I prefer keeping the function, as it makes it explicit that 0 doesn't have any special meaning besides being the default value.

src/bin/aftermath_convert_log.rs
andidr: I agree that although the check if a thread ID was discovered for an earlier `Trace` is performed quite frequently, it probably wouldn't have any significant impact on performance if we used simple set. However, the IDs are just plain integer values and we don't do anything special with them besides collecting and checking, so using a bit set from the start seems perfectly reasonable to me.

Unless you're worried about the extra dependency, I'd suggest to keep the bit set.

src/bin/aftermath_convert_log.rs
andidr: The scan_fmt! version is a bit more readable. However, If you're really worried about the extra dependency, I wouldn't mind changing the code as suggested.

src/offline_analysis/aftermath.rs
Elarnon: The point is that when you later try to copy or print one of those values it will be annoying to go back and add the derives. `Copy` is debatable but `Debug` and `Clone` are traits which usually just make sense to have on most (if not all) types.

src/bin/aftermath_convert_log.rs
Elarnon: Yes, I don't think adding an extra dependency for reasons of purity is worth it here.

src/bin/aftermath_convert_log.rs
Elarnon: I agree that the `scan_fmt!` version is more readable, but yes I think we should try not to add too much extra dependencies when it can easily be avoided.

src/offline_analysis/aftermath.rs
Elarnon: I guess technically you can do `d.map(duration_ns).unwrap_or_default()` then ;)
I'd suggest renaming the function `duration_ns_or_default` if you keep it; `duration_ns_default` looks like we are getting the "default duration in nanoseconds" (whatever that means).

andidr:
bors r=Elarnon r=ulysseB

andidr:
bors retry

andidr:
bors retry

andidr:
bors retry

PR #231, opened 2019-04-03
Add back variables test with fake device

Elarnon:
And other gardening funsies to make clippy happy and the warnings go
away.


nicoTolly:
Nothing special to say

Elarnon:
bors r=nicoTolly

PR #232, opened 2019-04-04
Small rewrite of the copy of the buffer in mppa backend

nicoTolly:
Separate casting from i8 to u8 from the copy to an array. As a minor bonus, copy now uses copy_from_slice which ought to be more efficient than iterating on a vector (in practice we don't care here).


nicoTolly:
bors  r=me

PR #234, opened 2019-04-04
Minor rewriting and renaming of NameMap

nicoTolly:
Title is quite explicit. Namer has been renamed ValuePrinter, Printer is now InstPrinter. name_inst becomes get_const_inst and no longer takes self (we don't need a trait object anymore). Backends have been rewritten accordingly. ValuePrinter is now an associated type of InstPrinter (Printer was generic over Namer before). As a result, X86Printer, CudaPrinter and MppaPrinter are no longer public. An alternative would have been to make each concrete implementation of ValuePrinter public, but I don't see the point of doing that. The possibility of further splitting ValuePrinter in ConstPrinter and Namer has been discussed but postponed as it does not offer any obvious immediate advantage.

### DIFF
--- a/backend/cuda/src/printer.rs
+++ b/backend/cuda/src/printer.rs
@@ -368,7 +368,7 @@ impl CudaPrinter {
 }
 
 impl InstPrinter for CudaPrinter {
-    type VP = ValuePrinter;
+    type ValuePrinter = ValuePrinter;
 
     /// Print result = op1 op op2
     fn print_binop(
--- a/backend/cuda/src/printer.rs
+++ b/backend/cuda/src/printer.rs
@@ -1,5 +1,5 @@
 //! Provides functions to print PTX code.
-use crate::{Gpu, Namer};
+use crate::{Gpu, ValuePrinter};
 use itertools::Itertools;
 use std::borrow::Cow;
 use std::fmt::Write as WriteFmt;
@@ -10,7 +10,7 @@ use telamon::search_space::{DimKind, Domain, InstFlag, MemSpace};
 use utils::*;
 
 #[derive(Default)]
-pub struct CudaPrinter {
+pub(crate) struct CudaPrinter {
     buffer: String,
 }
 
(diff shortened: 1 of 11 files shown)


backend/cuda/src/printer.rs
```diff
@@ -350,11 +361,8 @@ impl CudaPrinter {
     }
 }
 
-impl Printer<Namer> for CudaPrinter {
-    /// Get a proper string representation of an integer in target language
-    fn get_int(n: u32) -> String {
-        format!("{}", n)
-    }
+impl InstPrinter for CudaPrinter {
+    type VP = ValuePrinter;
```
Elarnon: I'd be explicit and call this `ValuePrinter`

backend/cuda/src/printer.rs
```diff
@@ -545,11 +553,11 @@ impl Printer<Namer> for CudaPrinter {
     fn name_operand<'a>(
         vector_levels: &[Vec<Dimension>; 2],
         op: &ir::Operand,
-        namer: &'a NameMap<Namer>,
+        name_map: &'a NameMap<ValuePrinter>,
```
Elarnon: shouldn't this be `NameMap<Self::ValuePrinter>`?  Here we use a trait object no?

Elarnon: Ah no this is trait impl not definition, ignore me.

src/codegen/name_map.rs
```diff
@@ -20,22 +20,22 @@ pub enum Operand<'a> {
 }
 
 /// Assign names to variables.
```
Elarnon: Should change this comment now

src/codegen/name_map.rs
```diff
@@ -435,9 +433,9 @@ impl VariableNames {
     }
 
     /// Creates the mapping from variables to names.
-    fn create(
+    fn create<VP: ValuePrinter>(
```
Elarnon: We probably could keep taking a trait object here to avoid template bloat

src/codegen/name_map.rs
```diff
@@ -446,15 +444,15 @@ impl VariableNames {
...
+                Self::new(var.t(), var.instantiation_dims(), value_printer)
             };
             vars_names.insert(var.id(), names);
         }
         vars_names
     }
 
     /// Creates a new set of names for a variable instantiated along the given dimensions.
-    fn new<IT>(t: ir::Type, dims: IT, namer: &mut Namer) -> Self
+    fn new<IT, VP: ValuePrinter>(t: ir::Type, dims: IT, value_printer: &mut VP) -> Self
```
Elarnon: Same as above, maybe keep it as trait object to avoid template bloat

src/codegen/printer.rs
```diff
@@ -26,9 +26,8 @@ impl MulMode {
 }
 
 #[allow(clippy::too_many_arguments)]
-pub trait Printer<N: Namer> {
-    /// Get the representation of an integer in the target language.
-    fn get_int(n: u32) -> String;
+pub trait InstPrinter {
+    type VP: ValuePrinter;
```
Elarnon: I'd be explicit and call this `type ValuePrinter: ValuePrinter;`

nicoTolly:
bors r=Elarnon

nicoTolly:
bors retry

nicoTolly:
bors retry

nicoTolly:
bors retry

nicoTolly:
bors retry

PR #236, opened 2019-04-05
Check results when exploring

Elarnon:
Currently we are only checking correctness in specific toy examples in
the tests; however, this is limited -- because we only run those tests
on CPU, they can miss backend-specific issues, but are also fully random
(we don't have a performance model to guide the search on CPU) which can
hide some issues which would happen in practice.

This patch makes it so that correctness tests are made during the
exploration as well, either for all new best implementations found (the
default) or for all implementations (for stress-testing).  When an
implementation fails the correctness check, we handle it the same as if
it had an infinite execution time -- which is technically incorrect but
should be good enough in practice for now.

The correctness checking is implemented by passing a checking function
as argument to the explorer APIs, which works even though it is kind of
ugly (those APIs are stating to grow an long list of arguments).  A
cleaner API would require some additional thought and refactoring, so I
leave it for future work.

Finally, there was an issue with the code in the cuda backend skipping
candidates for which the bound is higher than the current best:  since
the correctness check happens at a higher level, the backend thought it
had seen very fast implementations and using them to skip a bunch of
other implementations -- but the very fast implementations were actually
incorrect, and no longer used by the explorer.  Instead, I lifted the
skipping code to the explorer, resulting in an API change to the
`AsyncCallback` which now takes an `ExecutableKernel` as argument,
giving it control late in the pipeline to execute the kernel or not.  In
the same go I moved the stabilization from the CUDA crate to the
explorer as well, even though I realize when writing these lines that it
was not necessary.  Oh well, at least it is done now.

This patch also bumps itertools dependency to 0.8 everywhere, making all
crates depend on the same version.  This happens in this PR because the
stabilization code used itertools 0.8 (in the cuda backend crate) and
didn't work when moved to the main crate, which made me realize we
bundle *three* different versions of itertools in telamon.

Most of the changes correspond to simple updates to the new APIs (or to
some itertools changes);  the crux of the work happen in
`device/context.rs` and `explorer/mod.rs`.

Important changes:
	backend/cuda/:
		src/context.rs:  API update; remove stabilization
	backend/mppa/:
		src/context.rs:  API update; timing measurements fix
	backend/x86/:
		src/context.rs:  API update
	src/
		device/context.rs:  New `KernelEvaluator` and `Stabilizer` APIs
		explorer/candidate.rs: Allow creating with actions and dumping error data
		explorer/config.rs:  Check all or only best implementations.
		explorer/mcts.rs:  Rebuild the list of actions for the evaluator.
		explorer/mod.rs:  Check correctness and stabilize evaluations.


src/device/context.rs
```diff
@@ -88,7 +123,23 @@ impl<'a, T: ?Sized> ArgMapExt<'a> for T where T: ArgMap<'a> {}
 /// An evaluation context that runs kernels asynchronously on the target device.
 pub trait AsyncEvaluator<'a, 'b> {
     /// Add a kernel to evaluate.
-    fn add_kernel(&mut self, candidate: Candidate<'a>, callback: AsyncCallback<'a, 'b>);
+    fn add_any_kernel(
```
andidr: What is the idea behind the "any" in add_any_kernel?

src/device/context.rs
```diff
@@ -88,7 +123,23 @@ impl<'a, T: ?Sized> ArgMapExt<'a> for T where T: ArgMap<'a> {}
...
+        &mut self,
+        candidate: Candidate<'a>,
+        callback: AsyncCallback<'a, 'b>,
+    );
+}
+
+impl<'a, 'b, 'c> dyn AsyncEvaluator<'a, 'b> + 'c {
+    /// Helper to add a kernel to an evaluator trait object.  Using `add_any_kernel` directly trips
+    /// up type inference and requires explicitely typing closure arguments; using `add_kernel`
```
andidr: explicitly

src/device/context.rs
```diff
@@ -111,3 +162,151 @@ impl EvalMode {
         }
     }
 }
+
+#[derive(Debug, Clone)]
+pub struct Stabilizer {
```
andidr: A comment with a short description would be useful.

src/device/context.rs
```diff
@@ -111,3 +162,151 @@ impl EvalMode {
...
+    ) -> StableEvaluator<'a> {
+        StableEvaluator::new(self, kernel)
+    }
+
+    pub fn evaluate(&self, kernel: &mut dyn KernelEvaluator) -> Option<f64> {
+        self.wrap(kernel).evaluate()
+    }
+}
+
+pub struct StableEvaluator<'a> {
```
andidr: Short description

andidr:
> This patch also bumps itertools dependency to 0.8 everywhere, making all
> crates depend on the same version. This happens in this PR because the
> stabilization code used itertools 0.8 (in the cuda backend crate) and
> didn't work when moved to the main crate, which made me realize we
> bundle _three_ different versions of itertools in telamon.

This sounds like it should go into a separate commit.
 
> Most of the changes correspond to simple updates to the new APIs (or to
> some itertools changes); the crux of the work happen in
> `device/context.rs` and `explorer/mod.rs`.
> 
> Files changed:
> Cargo.toml:
> README.md:
> examples/matmul.rs:
> backend/cuda/:
> src/context.rs: API update; remove stabilization
> backend/mppa/:
> src/context.rs: API update; timing measurements fix
> backend/x86/:
> src/context.rs: API update
> kernels/:
> Cargo.toml:
> benches/cuda_variance.rs:
> src/kernel.rs:
> src/
> codegen/cfg.rs:
> device/context.rs: New `KernelEvaluator` and `Stabilizer` APIs
> device/fake.rs:
> device/mod.rs:
> explorer/candidate.rs: Allow creating with actions and dumping error data
> explorer/config.rs: Check all or only best implementations.
> explorer/mcts.rs: Rebuild the list of actions for the evaluator.
> explorer/mod.rs: Check correctness and stabilize evaluations.
> helper/builder.rs:
> ir/operator.rs:
> telamon-capi/:
> src/explorer.rs:
> telamon-gen/:
> Cargo.toml:
> cc_tests/Cargo.toml:
> src/ir/filter.rs:
> telamon-utils/
> Cargo.toml:
> src/iterator.rs:
> tests/
> common/mod.rs:

Maybe condense this by removing / grouping entries without a comment or replace with a higher-level overview, since the set of files with changes can already be easily determined using git.

src/device/context.rs
Elarnon: Because it takes a dynamic trait object, which can be any type implementing the proper trait.  I guess `add_dyn_kernel` would be better, I will change it.

Elarnon:
itertools bump could go in a separate commit but as explained in the commit message it is bundled because there are changes to the code using itertools as well.  As noted I don't think it is worth it to spend the time splitting in a separate commit (the changes are fairly minimal) especially since some code would have to be modified to work with new itertools version just to then be removed by the patch (or vice-versa).  I can split into a separate commit if you really think the extra work is worth it.

I added the info on the file list because @nicoTolly asked a couple month ago to include data about which file are modified for what and what important changes to look for in order to guide the reviewer.  I agree in this case it is a bit big (and also GitHub completely destroys my formatting :( ), but it does contains more info than what git has (since there are explanations of what is going on in each file).  Arguably I could remove files without a comment.

src/device/context.rs
Elarnon: Oops, thought I did that!

src/device/context.rs
```diff
@@ -111,3 +162,166 @@ impl EvalMode {
...
+        self.wrap(kernel).evaluate()
+    }
+}
+
+/// Stabilized evaluator as created by a `Stabilizer`.
+///
+/// This wraps a `KernelEvaluator` with a `Stabilizer`, which is used to configure the number of
+/// runs to do when averaging evaluation.
+///
+/// The stable evaluator can also contains a lower bound on the runtime and the best runtime seen
```
andidr: contain

src/device/context.rs
andidr: Not quite sure about the naming. "Stabilizer" sounds more like something that compensates a malfunction.

andidr:
> itertools bump could go in a separate commit but as explained in the commit message it is bundled because there are changes to the code using itertools as well. As noted I don't think it is worth it to spend the time splitting in a separate commit (the changes are fairly minimal) especially since some code would have to be modified to work with new itertools version just to then be removed by the patch (or vice-versa). I can split into a separate commit if you really think the extra work is worth it.

If this can be done with minimal effort, I would prefer putting it into a separate commit.

> I added the info on the file list because @nicoTolly asked a couple month ago to include data about which file are modified for what and what important changes to look for in order to guide the reviewer. I agree in this case it is a bit big (and also GitHub completely destroys my formatting :( ), but it does contains more info than what git has (since there are explanations of what is going on in each file). Arguably I could remove files without a comment.

I would just leave the files with comments and skip the rest. However, it's probably up to @nicoTolly to decide on this.

Elarnon:
bors r=andidr

I ended upsplitting the itertools bump in a separate commit.

PR #237, opened 2019-04-05
Use relative error in kernel error checking

Elarnon:
Currently we are checking kernel output using a single threshold of
`1e-5` on the absolute error.  However, for large kernels, this is not
appropriate: errors accumulate on large sums, and can overflow the
threshold even for correct computations.

Instead, this patch makes it so that we check correctness using a
combination of an absolute and relative threshold.  This alleviates the
problem since large sums will have higher absolute values[1] and hence
will allow for more leeway thanks to the relative threshold term.

The implementation is similar to the NumPy `np.allclose` function, that
is, we add the absolute and relative tolerance and compare with the
absolute difference.  The absolute and relative tolerance are 1e-8 and
1e-5 respectively, following again the default values from
`np.allclose`.

At the same time, the patch also changes the error displayed in case of
correctness errors:  instead of dumping sometimes large tensors in
a textual form, it now prints summary statistics of the absolute and
relative errors, as well as the proportion of elements failing the
`allclose` check.

Note that those two issues are related to now checking correctness of
full kernels during evaluation: where previously we were only checking
correctness on small toy examples in tests, we are now checking
correctness on large tensors -- which makes the error messages way too
large (we search axpy with sizes of 2^26 -- one single error message was
gigabytes long!), and also introduces the additional precision issues.

Additionally, one should note that, due to the transformations we
currently support, there should not be any error at all (in fact, if one
looks closely, the errors for all correct Telamon kernels are always
the same).  In particular, we do not reorder the arithmetic operations.
The error that we observe should thus be attributed to (a) optimizations in
the reference implementations we compare against compared to the naive
version and possibly (b) the use of fused multiply-add instructions on
some platforms.

 1: Except where loss of significance occurs, typically.  But since we
    only generate positive random floats this shouldn't be an issue for
    us in practice; if it is, we can revisit the approach and e.g. scale the
    error depending on the kernel sizes.  The advantage of using a
    relative error is that it doesn't require any inner knowledge of the
    kernel semantics and provide a generally acceptable check.

kernels/:
 - src/lib.rs:  Add new error-checking functions
 - src/linalg.rs:  Use the new error-checking functions
 - src/kernel.rs:  Add missing space.

### DIFF
--- a/kernels/src/lib.rs
+++ b/kernels/src/lib.rs
@@ -6,12 +6,16 @@ mod kernel;
 pub mod linalg;
 pub mod statistics;
 
+use std::fmt;
+
 pub use crate::kernel::{analyze_bounds, Kernel};
 
 use telamon::device::{self, ArgMap, Context};
 use telamon::helper::tensor::DimSize;
 use telamon::helper::{self, SignatureBuilder};
 use telamon::{explorer, model, search_space};
+
+use ::ndarray::{ArrayBase, Data, Dimension, FoldWhile, Zip};
 
 /// Creates a candidate from the search space and registers the tile sizes in it.
 fn build_candidate<'a>(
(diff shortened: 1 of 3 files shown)


andidr:
> [...]
> Currently we are checking kernel output using a single threshold of
> `1e-5` on the absolute error. However, for large kernels, this is not
> appropriate: errors accumulate on large sums, and can overflow the
> threshold even for correct computations.

Replace "overflow" with "exceeded" ("overflow" sounds as if the maximum value of the data type for the error was exceeded).

> [...]
> Instead, this patch makes it so that we check correctness using a
> combination of an absolute and relative threshold. This alleviates the
> problem since large sums will have higher absolute values[1] and hence

Not essential, but the [1] looks more like a reference than a footnote. Maybe just use an asterisk?

kernels/src/lib.rs
```diff
@@ -51,35 +55,183 @@ fn infer_tiling(
...
+                FoldWhile::Continue(())
+            } else {
+                FoldWhile::Done(())
+            }
+        })
+        .is_done()
+}
+
+#[derive(Debug)]
+struct IncorrectOutput<S> {
```
andidr: Documentation missing

Elarnon:
bors retry

PR #239, opened 2019-04-10
Introduce a simpler API for building kernels

Elarnon:
This patch introduces `KernelBuilder`, which is a tiny wrapper around
`SignatureBuilder` for building kernels, simplifying the creation of a
kernel, especially when creating the kernel outside the `Kernel` trait.

In addition, it makes the memory initialization strategy explicit using
an appropriate `MemInit` enum instead of using naked booleans.


Elarnon:
bors r=nicoTolly,andidr

PR #240, opened 2019-04-12
Minor Fixes in mppa

nicoTolly:
Two minor errors - a superfluous semi-column and a typo - were introduced by mistakes - and also the fact that we can't compile telamon-mppa on standard machines.


nicoTolly:
bors r=me

PR #241, opened 2019-04-12
Adapt Mppa backend to lifetime removal

nicoTolly:
Context Trait implementation of mppa was not in par with current master version


nicoTolly:
bors r=me

PR #242, opened 2019-04-12
Fake implementation of telajax

nicoTolly:
This PR adds a fake implementation of telajax that allows the code to be compiled on every architecture. Mppa syntax checking has been added in travis.yaml too.

### DIFF
--- a/backend/mppa/src/context.rs
+++ b/backend/mppa/src/context.rs
@@ -11,7 +11,10 @@ use std::sync::{
 };
 use std::time::Instant;
 use std::{self, fmt};
+#[cfg(feature = "real_mppa")]
 use telajax;
+#[cfg(not(feature = "real_mppa"))]
+use crate::fake_telajax as telajax;
 use telamon::codegen::{Function, NameMap, ParamVal};
 use telamon::device::{
     self, ArrayArgument, AsyncCallback, Context as ContextTrait, EvalMode,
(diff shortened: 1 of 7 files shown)


backend/mppa/src/context.rs
```diff
@@ -1,4 +1,6 @@
 //! MPPA evaluation context.
+#[cfg(not(feature = "real_mppa"))]
+use crate::fake_telajax as telajax;
```
Elarnon: I would put this one close to the real `use telajax`

backend/mppa/src/printer.rs
```diff
@@ -222,8 +222,8 @@ impl MppaPrinter {
         for i in 0..n {
             let start = format!("d{}", i);
             let mut vec_str = vec![start];
-            for j in 0..i {
-                vec_str.push(format!("{}", unwrap!(dims[j].size().as_int())));
+            for dim in dims.iter().take(i) {
```
Elarnon: I see you ran clippy :)

nicoTolly: Precisely :)

backend/mppa/src/context.rs
nicoTolly: Oh yes, cargo fmt did that for me, I think

nicoTolly: No choice here, if I change that cargo fmt -- --check will fail... Is there any workaround ?

nicoTolly:
bors r=Elarnon

PR #244, opened 2019-04-17
Removed unnamed types in telajax and fixed a typo

nicoTolly:
*const _ is indeed a dangerous type. At some point we used a &self.inner where it should have been a &*self.inner (as self.inner is an Arc, it must be dereferenced before taking a pointer on the enclosed data). This error was silenced because of the cast to an unnamed type. Consequently, all cast of this kind have been removed from the code (and typo was fixed).


nicoTolly:
bors r=me

PR #245, opened 2019-04-19
Remove old bandit arm implementation

Elarnon:
This was kept in case issues were found with the new MCTS
implementation.  It has now been extensively used for over a month
without issues and with better performance due to the more aggressive
caching it does, so there are no real reasons to keep maintaining
`bandit_arm` anymore.


Elarnon:
bors retry

Elarnon:
bors retry

Elarnon:
bors retry

Elarnon:
bors retry

PR #246, opened 2019-04-24
Add options --exclude-traces and --exclude-evaluations to aftermath_convert_log

andidr:
This patch adds two options to aftermath_convert_log, allowing the user to
reduce the size of the output trace.

The option --exclude-traces disables conversion of log events originating from
Message::Trace messages, while --exclude-evaluations disables conversion of
Message::Evaluation messages.

If both options are specified at the same time, a minimal Aftermath trace
containing only the candidate tree is written.


andidr:
bors r=elarnon

PR #247, opened 2019-04-26
Format floating point literals correctly in x86 and MPPA backends

andidr:
Currently, both the x86 and MPPA backend print floating point literals using
f64::to_string(). This causes two problems.

First, floating point values without decimal places are printed without a
trailing ".0", which lets the C compiler interpret them as integer
literals. This results in a warning for valid floating point values which are
outside of the range of an integer.

Second, this prevents the C compiler from emitting a warning for values that
should be represented as C floats, but which are out of range.

This patch adds proper suffixes to floating point literals, avoiding the
above-mentioned problems. For floating point types with an unusual number of
bits (i.e., neither 32 nor 64 bits), the formatting function panics.

### DIFF
--- a/backend/mppa/src/lib.rs
+++ b/backend/mppa/src/lib.rs
@@ -15,6 +15,7 @@ pub use crate::mppa::Mppa;
 
 use num::bigint::BigInt;
 use num::rational::Ratio;
+use num::traits::Float;
 use num::ToPrimitive;
 use telamon::{codegen, ir};
 use utils::*;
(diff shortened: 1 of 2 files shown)


andidr:
> Didn't think about it yesterday but I think we can use scientific notation to force parsing as float on the C side:
I finally switched to hexadecimal floating point literals introduced by C99, which also correctly handles values which cannot be represented in decimal notation.

andidr:
bors retry

PR #249, opened 2019-05-01
Remove BoundGuard in offline analysis

Elarnon:
Somehow I never knew that `std::cell::Ref` (and `std::cell::RefMut`)
handled partial borrows [1].

 1: https://doc.rust-lang.org/std/cell/struct.Ref.html#method.map


Elarnon:
bors r=andidr

PR #250, opened 2019-05-03
Add kernel for fused matrix multiplication with activation function

andidr:
Add a new kernel FusedMatMulAct for fused matrix multiplication with activation function (Identity, ReLU or Sigmoid). The implementation required adding two new instructions: max and exp.

The existing MatMul kernel has been remodeled as a special case of the fused version with the Identity function as the activation function.

Tested with the CUDA and x86 backends.


backend/cuda/src/printer.rs
```diff
@@ -402,8 +403,33 @@ impl InstPrinter for CudaPrinter {
...
+                    ir::Type::F(32) => {
+                        // Move constant for log2(e) into register
+                        unwrap!(writeln!(
+                            self.buffer,
+                            "mul.{t} {reg}, 0f3fb8aa3b, {operand}; // 0f3fb8aa3b = log2(e)\n",
+                            t = t_str,
+                            reg = result,
+                            operand = operand
+                        ));
+                        (format!("ex2.approx"), result)
```
Elarnon: Thinking about this more -- I think it'd be clearer to write everything for the `Exp` case in a single match branch and replicate the `writeln!` for the two others (with a helper function if really needed).  Currently the actualy generated code for `Exp` is hard to follow due to consisting of two instructions which are printed somewhat far apart.

backend/x86/src/printer.rs
```diff
@@ -411,12 +422,19 @@ impl InstPrinter for X86printer {
...
+            }
+
             ir::UnaryOp::Cast(t) => {
-                unwrap!(write!(self.buffer, "({})", Self::get_type(t)))
+                unwrap!(write!(self.buffer, "({}){};", Self::get_type(t), operand));
             }
+
+            ir::UnaryOp::Exp(t) => match t {
+                ir::Type::F(32) => unwrap!(write!(self.buffer, "expf({});", operand)),
+                _ => panic!("Exp not implemented for given type"),
```
Elarnon: This should print out said "given type"

kernels/src/linalg.rs
```diff
@@ -365,115 +362,34 @@ impl<'a, S: Scalar> Kernel<'a> for MatMul<'a, S> {
...
-        let a = TensorBuilder::new("a", a_dims)
-            .doif(params.transpose_a, |b| b.transpose(0, 1))
-            .stride_dim(2)
-            .finish(builder);
-        let b = TensorBuilder::new("b", vec![k_size, n_size.clone()])
-            .doif(params.transpose_b, |b| b.transpose(0, 1))
-            .finish(builder);
-        let c = builder.tensor::<S>("c", vec![m_size, n_size], false);
-        MatMul { params, a, b, c }
+        let fmma_params = FusedMatMulActP {
```
Elarnon: Probably should implement `From<MatMulP> for FusedMatMulActP` and use `params.into()`

kernels/src/linalg.rs
```diff
@@ -658,3 +574,206 @@ impl<'a, S: Scalar> Kernel<'a> for BatchMM<'a, S> {
...
+pub struct FusedMatMulAct<'a, S: Scalar> {
+    pub params: FusedMatMulActP,
+    a: Tensor<'a, S>,
+    b: Tensor<'a, S>,
+    c: Tensor<'a, S>,
+}
+
+#[derive(Clone, Deserialize, Serialize)]
+pub enum ActivationFunction {
+    /// Activation function that leaves the original value (i.e., nop)
```
Elarnon: I'd suggest not using an inline constructor for this but rather an `Option`.  It would make a 1-1 mapping between "there is an activation function -> we have an extra instruction + loops / there is no activation -> we don't have the extra instruction at all"

Elarnon: (Also nitpicking but this type could simply be called `Activation`)

kernels/src/linalg.rs
```diff
@@ -658,3 +574,206 @@ impl<'a, S: Scalar> Kernel<'a> for BatchMM<'a, S> {
...
+        let a_op = ld_a.dim_map(&[&acc_dim_m, &acc_dim_k], GlobalScope(()), &mut builder);
+        let b_op = ld_b.dim_map(&[&acc_dim_k, &acc_dim_n], GlobalScope(()), &mut builder);
+        let acc = builder.mad(&a_op, &b_op, &helper::Reduce(acc_init));
+
+        let act = match self.params.activation_fun {
+            ActivationFunction::ReLU => builder.max(&S::from(0.0).unwrap(), &acc),
+            ActivationFunction::Sigmoid => {
+                let exp = builder.exp(&acc);
+                let add = builder.add(&S::from(1.0).unwrap(), &exp);
+                builder.div(&S::from(1.0).unwrap(), &add)
```
Elarnon: I am somewhat surprised that this works given that you didn't close the `acc_dim_k` dimension before

kernels/src/linalg.rs
```diff
@@ -658,3 +574,206 @@ impl<'a, S: Scalar> Kernel<'a> for BatchMM<'a, S> {
...
+            self.params.mm_params.k as usize,
+            self.params.mm_params.n as usize,
+        );
+        let a = unwrap!(self.a.read_to_host(context).into_shape(a_shape));
+        let b = unwrap!(self.b.read_to_host(context).into_shape(b_shape));
+        let mut res = a.dot(&b);
+
+        match self.params.activation_fun {
+            ActivationFunction::ReLU => {
+                let zero = S::from(0).unwrap();
```
Elarnon: `S::zero()` should work I think (if it doesn't we should add the appropriate `: Zero` bound in `ScalarArgument`)

kernels/src/linalg.rs
```diff
@@ -658,3 +574,206 @@ impl<'a, S: Scalar> Kernel<'a> for BatchMM<'a, S> {
...
+
+                for c in res.iter_mut() {
```
Elarnon: I guess `res.mapv_inplace(|c| c.max(zero))` would be more idiomatic

kernels/src/linalg.rs
```diff
@@ -658,3 +574,206 @@ impl<'a, S: Scalar> Kernel<'a> for BatchMM<'a, S> {
...
+                    if *c < zero {
+                        *c = zero;
+                    }
+                }
+            }
+
+            ActivationFunction::Sigmoid => {
+                let one = S::from(1.0).unwrap();
```
Elarnon: Same as above; use `S::one()`

kernels/src/linalg.rs
```diff
@@ -658,3 +574,206 @@ impl<'a, S: Scalar> Kernel<'a> for BatchMM<'a, S> {
...
+
+                for c in res.iter_mut() {
```
Elarnon: Again I guess `res.mapv_inplace(|c| one / (one + S::exp(c)))` is more idiomatic

Elarnon: (and possibly faster since there are less ordering constraints for `mapv_inplace` and `ndarray` could parallelise the calls, although it doesn't for now afaik)

kernels/src/linalg.rs
andidr: MatMulP only defines how the matrix multiplication should be done and choosing the identity function as the default activation function would be arbitrary. However, from the perspective of the MatMul kernel, this choice well justified and should be explicit, just as it is now.

kernels/src/linalg.rs
andidr: > I'd suggest not using an inline constructor for this but rather an `Option`. It would make a 1-1 mapping between "there is an activation function -> we have an extra instruction + loops / there is no activation -> we don't have the extra instruction at all"

You could argue that conceptually, there is always an activation function and that, by chance, one of them can be implemented with zero instructions. A None value for an Option would also make this a little weird: you'd get "a fused matrix multiplication with an activation function without an activation function". As there is zero overhead for the identity function, I prefer keeping things the way they are. 

> (Also nitpicking but this type could simply be called `Activation`)

Activation sounds as if this would refer to the process of applying the activation function. I'd rather keep ActivationFunction to match the usual terminology and make clear that this identifies the function that is applied.

kernels/src/linalg.rs
Elarnon: I don't understand what you are saying.  `MatMulP` defines a matrix multiplication, `FusedMatMulP` defines a matrix multiplication followed by an identity function.  Why wouldn't it make sense to convert from one to the other?

BTW "from the perspective of the MatMul kernel" vs "from the perspective of MatMulP" doesn't really make sense to me either.  The way I see it, the only difference is that `MatMulP` is a non-instantiated matrix multiplication kernel while `MatMul` is an instantiated matrix multiplication kernel (with parameters allocated on some device and code generated).

kernels/src/linalg.rs
Elarnon: >You could argue that conceptually, there is always an activation function and that, by chance, one of them can be implemented with zero instructions.

It's true that this is less confusing given your current implementation; in my mind, the fused matrix multiply would add loops as well, i.e. `C = f(A x B)` is implemented as (ignoring tiling):

```
%0[@0, @1] = load(A[@0, @1])
%1[@2, @3] = load(B)
%2[@4, @5, @6] = mov(0)
%3[@7, @8, @9] = fma(%0[@7, @8], %1[@8, @9], reduce(%2, [@8]))
%4[@10, @11] = f(%3[@10, Last, @11])
%5[@12, @13] = store(C, %4[@12, @13])
```

and I would argue that a fused matrix multiply with identity activation would use this code structure, (with `f` being a no-op)  which is different from the code structure of regular matrix multiplication.

For comparison your current implementation is this, where some merging choices have already been made (note that the same dimensions are used for `%4` and `%3`):

```
%0[@0, @1] = load(A[@0, @1])
%1[@2, @3] = load(B)
%2[@4, @5, @6] = mov(0)
%3[@7, @8, @9] = fma(%0[@7, @8], %1[@8, @9], reduce(%2, [@8]))
%4[@7, @9] = f(%3[@7, @9])
%5[@10, @11] = store(C, %4[@10, Last, @11])
```
Thinking about it more, I'm not sure I agree with this implementation instead of the first one; let's discuss on Monday.

> A None value for an Option would also make this a little weird: you'd get "a fused matrix multiplication with an activation function without an activation function".

In my view, there is no "fused matrix multiplication with an activation function"; there is "matrix multiplication which supports an optional activation function".  So I would have *both* `None` and `Some(ActivationFn::Identity)` be valid, but do different things :)

kernels/src/linalg.rs
andidr: Apologies if that wasn't clear. By implementing the From trait, there will be a generic conversion from MatMulP to FusedMatMulP. Hence, every time a MatMulP is used to initialize the parameters of the fused version, the identity function will be used as the activation function. While this makes sense for the simple matrix multiplication, it might not make sense is another context.

I would prefer leaving the selection of the activation function as something that needs to be stated explicitly, which is how things are currently implemented.

backend/mppa/src/printer.rs
```diff
@@ -474,21 +474,32 @@ impl InstPrinter for MppaPrinter {
...
+            ir::BinOp::Equals => ("==", true),
+            ir::BinOp::Max => ("telamon_op_max", false),
         };
-        unwrap!(writeln!(
-            self.buffer,
-            "{} = {} {} {};",
-            result, lhs, op, rhs
-        ));
+
+        if is_infix_op {
```
nicoTolly: Compilation error here : you should either use is_infix or rename is_infix to is_infix_op (which would be coherent with what you did in x86)

andidr:
Pushed a new version:
- Moved instructions of the activation function to new set of nested dimensions
- Renamed FusedMatMulAct to FusedMM
- Replaced MatMul entirely with FusedMM
- Added Option for the activation function with None -> Identity Function
- Added tests for all activation functions

andidr:
bors retry

andidr:
bors r=Elarnon,nicoTolly

andidr:
bors retry

PR #253, opened 2019-05-09
Improve internal data structures pretty-printing

Elarnon:
This patch improves the display of various internal Telamon structures.
Notably:

 - Actions are displayed in the format `DimKind(%2) = {BLOCK, LOOP}` and
   `DimKind(%2) = {BLOCK}` instead of `DimKind(%2, BLOCK | LOOP)` and
   `DimKind(%2, BLOCK)`.  This is closer to the formal definitions, and
   better defined since there is a clear separation between the choice
   and domain.

 - Numeric domains (i.e. sizes) are displayed using their actual values
   instead of their index in the bitset vector.  For instance, if a size
   has possible values {2, 4, 16}, before we would display each of those
   values as `NS { 1 }`, `NS { 2 }` and `NS { 4 }` respectively -- now
   we properly display `{2}`, `{4}` and `{16}`.

 - `codegen::Cfg` objects are now printed using full instruction
   information and should contain enough data to fully understand the
   code behavior without external information (eg dimension mappings).

 - `codegen::Function` can now be printed, which will print the `Cfg`
   but also additional information such as the size of block and thread
   dimensions.

 - Most of the `ir` types have a more precise `Display` information
   without relying too much on objects IDs, which tries to strike a
   balance between verbosity and usefulness.

   Most of those types need an ir::Function to be properly displayed,
   since they need to resolve IDs to get the required information.  We
   use a new trait, `ir::IrDisplay`, which is similar to `fmt::Display`
   but requires an additional reference to an `ir::Function` to be
   passed in.  To integrate with the usual Rust formatting pipelines, a
   `.display()` method is provided, taking as argument the reference to
   the `ir::Function`, and returing a wrapper object which passes it
   back to `ir::IrDisplay::fmt` when displayed.  This allows the
   following pattern:

   ```rust
   let obj = ...; // Some object implementing IrDisplay
   let fun = ...; // An ir::Function instance
   println!("{}", obj.display(&fun));
   ```

### DIFF
--- a/src/codegen/cfg.rs
+++ b/src/codegen/cfg.rs
@@ -1,7 +1,7 @@
 use crate::codegen::{Dimension, InductionLevel, Instruction};
 use crate::ir;
 use crate::search_space::*;
-use itertools::{self, Itertools};
+use itertools::Itertools;
 use log::debug;
 use std::{self, fmt};
 use utils::unwrap;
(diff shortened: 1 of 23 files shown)


Elarnon:
bors try

Elarnon:
bors try

src/codegen/cfg.rs
```diff
@@ -204,6 +204,140 @@ impl<'a> Cfg<'a> {
...
+            on_newline: true,
+        }
+    }
+}
+
+impl fmt::Write for IndentAdapter<'_> {
+    fn write_str(&mut self, mut s: &str) -> fmt::Result {
+        while !s.is_empty() {
+            if self.on_newline {
+                self.fmt.write_str("  ")?;
```
andidr: The IndentAdapter might be useful for pretty printing elsewhere in the project and it might make sense to make the indent configurable

src/ir/dimension.rs
```diff
@@ -243,6 +249,12 @@ impl<L> Statement<L> for Dimension<L> {
 #[repr(transparent)]
 pub struct LogicalDimId(pub u32);
 
+impl fmt::Display for LogicalDimId {
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        write!(fmt, "~{}", self.0)
```
andidr: Maybe use a more recognizable prefix, e.g., L instead of ~ (or ~L)

src/codegen/cfg.rs
Elarnon: Agreed, but let's leave this to the person actually reusing the `IndentAdapter`.

src/ir/dimension.rs
Elarnon: We use a single character sigil for other objects, so `~L` would be inconsistent.  I also explicitely don't want to use letters to make sure that the displayed values can always be distinguished from valid identifiers, since they can appear in the same context as identifiers (e.g. in action args).

Elarnon:
bors r=andidr

PR #254, opened 2019-05-09
Variable name for closed dims refer to the last iteration

Elarnon:
The name map incorrectly assumed that it should use the *first*
iteration index instead of the *last* one when some unrolling dimensions
are not used by the underlying instruction (eg because it has been
replicated).  This makes us generate invalid code for reductions,
because it causes the reduction operand to get out of sync, generating
code similar to the following:

```
x0 = 0
x1 = 0
x0 = a0 * b0 + x1
x0 = a1 * b1 + x1
```

instead of

```
x0 = 0
x1 = 0
x1 = a0 * b0 + x1
x1 = a1 * b1 + x1
```

This is a partial fix for the same issue as #226.  We are still creating
too many variables when unrolled loops are unrolled, but at least the
generated code should be correct.

### DIFF
--- a/src/codegen/name_map.rs
+++ b/src/codegen/name_map.rs
@@ -430,7 +430,9 @@ impl VariableNames {
             .iter()
             .zip_eq(&self.names.dims)
             .map(|(index, size)| match index {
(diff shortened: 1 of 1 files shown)


Elarnon:
bors retry

PR #258, opened 2019-05-13
Add test for reproducible code generation by the MPPA backend

andidr:
Add a new test that checks for a set of implementations if the generated code is identical to the code generated when the implementations were found.


andidr:
bors r=Elarnon r=nicoTolly

Elarnon:
This shouldn't be r=me; I did not look at it (not an issue for merging, but we should try to have commits indicate who actually reviewed them).

PR #259, opened 2019-05-15
Use hash maps from fxhash instead of fnv

andidr:
Replace the hash maps from `fnv` with implementations from `fxhash`, resulting in a ~30% decrease of execution time of the `descent` bench.


Cargo.toml
```diff
@@ -12,10 +12,6 @@ repository = "https://github.com/ulysseB/telamon/"
 name = "descent"
 harness = false
 
-[[bench]]
```
Elarnon: Revert this

andidr: That shouldn't have been committed. Fixed in the update.

andidr:
bors r=Elarnon

andidr:
bors retry

PR #261, opened 2019-06-11
Add EXH mode for Emacs

andidr:
Add an Emacs major mode `exh-mode' with syntax coloring and
indentation for EXH files.


andidr:
bors r=Elarnon

PR #262, opened 2019-06-14
Proper into_num_set() default impl

Elarnon:
The default implementation for into_num_set() was erroneously ignoring
the last available value.  In particular it would transform a singleton
set into the empty set.  This makes it so that the last value is
properly accounted for.


Elarnon:
bors r=andidr

PR #263, opened 2019-06-14
Rename `DimMap::filter` into `DimMap::drain_filter`

Elarnon:
`DimMap::filter` is a confusing name: not only does it return an
iterator over the filtered elements, but it also removes them from the
`DimMap`.  This can make for very surprising behavior, since `.filter()`
(e.g. on an iterator) usually does not have side effects.

The behavior and intent is (almost) identical to the nightly
`drain_filter` on `Vec`, so let's use that name instead to be less
confusing and more explicit.

### DIFF
--- a/src/ir/dim_map.rs
+++ b/src/ir/dim_map.rs
@@ -11,9 +11,9 @@ use utils::*;
 
 /// Represents a mapping between dimenions.
 #[derive(Clone, Debug)]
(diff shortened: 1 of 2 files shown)


Elarnon:
bors r=andidr

Elarnon:
bors retry

PR #266, opened 2019-07-09
Allow creating explorer config from any file

Elarnon:
Previously explorer configuration could only be created from
`Settings.toml` files, or a json string.  This patch makes it so that
explorer configuration can be created by any file or json string as
needed by the caller.

In addition, we push errors to the creator of the `Config` instead of
unwrapping, and we output a warning message when automatically loading
from a `Settings.toml` since this is essentially context-dependent
defaults which can be confusing.

### DIFF
--- a/src/explorer/config.rs
+++ b/src/explorer/config.rs
@@ -10,6 +10,7 @@ use std::path::{Path, PathBuf};
 use std::{self, error, fmt, str::FromStr};
 
 use config;
+use log::warn;
 use num_cpus;
 use serde::{Deserialize, Serialize};
 use utils::{tfrecord, unwrap};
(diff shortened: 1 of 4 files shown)


src/explorer/config.rs
```diff
@@ -60,24 +61,35 @@ impl Config {
...
-            unwrap!(config_parser.merge(config::File::from(config_path)));
-        }
         config_parser
     }
 
-    /// Extract the configuration from the configuration file, if any.
-    pub fn read_from_file() -> Self {
-        unwrap!(Self::create_parser().try_into::<Self>())
+    /// Create a new configuration from the hardcoded `Settings.toml` configuration file if it
+    /// exisst.
```
andidr: exists

Elarnon:
sure, thanks.

Elarnon:
bors r=andidr

PR #267, opened 2019-07-10
Add kernel for ResNet cell

andidr:
Add a new kernel `ResNetCell` implementing one cell of ResNet [1]. The kernel computes:

  O = activation(activation(A.B).C) + A

where `A`, `B` and `C` are matrices and `activation` an activation
function (ReLU, Sigmoid or identity).

[1] https://arxiv.org/abs/1512.03385

### DIFF
--- a/kernels/src/linalg.rs
+++ b/kernels/src/linalg.rs
@@ -3,8 +3,8 @@
 use std::sync::Arc;
 
 use crate::compose::{
-    matrix_matrix_multiply, matrix_vector_multiply, tensor_elementwise_mul, tensor_mad,
-    ActivationFunction,
+    array_activate_inplace, matrix_matrix_multiply, matrix_vector_multiply,
+    tensor_activate, tensor_add, tensor_elementwise_mul, tensor_mad, ActivationFunction,
 };
 use crate::kernel::Kernel;
 use crate::{build_candidate, check_output, create_size, infer_tiling, Scalar};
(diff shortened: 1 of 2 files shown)


kernels/src/linalg.rs
```diff
@@ -835,3 +807,182 @@ impl<'a, S: Scalar> Kernel<'a> for Fused2MM<'a, S> {
...
+    }
+
+    fn build_body<'b>(
+        &self,
+        signature: Arc<ir::Signature>,
+        ctx: &'b dyn device::Context,
+    ) -> Vec<Candidate> {
+        let m_tiling = infer_tiling(self.params.m, &self.params.m_tiling, &[32, 4]);
+        let n_tiling = infer_tiling(self.params.n, &self.params.n_tiling, &[32, 4]);
+        let k_tiling = infer_tiling(self.params.k, &self.params.k_tiling, &[32]);
```
Elarnon: I think we want `32, 4` here as well?  Usually we don't b/c the `k` dimension is only inside the matrix mutiply but here it also is an output dimension.

kernels/src/linalg.rs
```diff
@@ -835,3 +807,182 @@ impl<'a, S: Scalar> Kernel<'a> for Fused2MM<'a, S> {
...
+            .c
+            .load(vec![n_tiling.clone(), k_tiling.clone()], &mut builder);
+
+        let ab = matrix_matrix_multiply(&mut builder, &a, &b);
+        let act_ab = tensor_activate::<S>(&mut builder, ab, &self.params.activation_fun);
+        let act_ab_c = matrix_matrix_multiply(&mut builder, &act_ab, &c);
+        let act_act_ab_c =
+            tensor_activate::<S>(&mut builder, act_ab_c, &self.params.activation_fun);
+
+        let a_copy = self
```
Elarnon: Why do we need to load `a` again?  (I see various possible technical answers to this... but I don't think there are theoretical reasons to do so, unless I am mistaken)

kernels/src/linalg.rs
```diff
@@ -835,3 +807,182 @@ impl<'a, S: Scalar> Kernel<'a> for Fused2MM<'a, S> {
...
+            .a
+            .load(vec![m_tiling.clone(), k_tiling.clone()], &mut builder);
+
+        let act_act_ab_c_pa = tensor_add(&mut builder, &act_act_ab_c, &a_copy);
```
Elarnon: As an aside -- we probably need to have a way to remap dimensions here to "clean things up" and avoir all the useless dimensions in this addition.  This should probably come "for free" if we implement convolutions, as we need a more complex version of the same thing iirc.

PR #268, opened 2019-07-12
Add `threads` and `thread_size` choices groups

Elarnon:
- The `threads` group generate choices for determining if each
   dimension should be implemented as a thread or not.

 - The `thread_size` group fixes the sizes only of the dimensions which
   may be mapped to threads, which has a large impact on the mem_model
   analysis for access conflicts and replays.

Some quality of life changes are also included:

 - The custom `NestedIterator` implementation is removed, as it turns
   out the standard library already provides that construct through
   `.flatten()`

 - When applying an action fails, `ActionEx`-level APIs now return a
   proper error instead of `()`, which provides actually useful error
   messages.


Elarnon:
bors r=andidr

PR #272, opened 2019-07-24
Don't update the cut if it is worse than before

Elarnon:
When using the `initial_cut` parameter (eg when reusing a cut found from
a previous search), we were updating the cut after a first valid
implementation is found.  Such an implementation would have a bound that
is less than the initial cut, but not necessarily (in fact, almost
never) an execution time lower than the initial cut -- thus we would in
fact make the cut worse by performing this update.

This patch instead changes it so that we only update the cut if it
improves it.

NB: An alternative fix for this issue would be to take the initial cut
into account in the monitor, but this means we would ignore
implementations that are not faster than the initial cut.  This would
limit the usefulness of the `initial_cut` parameter -- in particular, it
wouldn't be possible to use it to optimisticaly cut branches with an
aggressive cut for which we don't have an implementation.

### DIFF
--- a/src/explorer/mcts.rs
+++ b/src/explorer/mcts.rs
@@ -1266,7 +1266,14 @@ where
     type Event = Message;
 
     fn update_cut(&self, new_cut: f64) {
-        *self.cut.write().expect("cut: poisoned") = new_cut;
+        // If an initial cut was specified in the configuration file, `update_cut` will be called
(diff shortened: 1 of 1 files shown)


Elarnon:
bors r=andidr

Elarnon:
bors retry

Elarnon:
bors retry

PR #273, opened 2019-07-26
Add choice groups for fusion and nesting of dimensions

andidr:
Add new two new choice groups `DimFusion` and `DimNesting` that allow for the separation of fusion and nesting choices from sequential ordering choices from `Order`.

### DIFF
--- a/src/explorer/choice.rs
+++ b/src/explorer/choice.rs
@@ -125,6 +125,25 @@ pub fn list<'a>(
                         )
                     }))
                 }
+
+                ChoiceGroup::DimFusion => {
+                    Box::new(fun.dims().enumerate().flat_map(move |(i, lhs)| {
+                        let lhs = lhs.stmt_id();
+                        let dims = fun.dims().take(i).map(|x| x.stmt_id());
+
+                        dims.chain(fun.insts().map(|x| x.stmt_id())).flat_map(
+                            move |rhs| {
+                                let available_orders = space.domain().get_order(lhs, rhs);
+
+                                gen_choice(
+                                    available_orders.bisect(Order::MERGED),
+                                    &|order| Action::Order(lhs, rhs, order),
+                                )
+                            },
+                        )
+                    }))
+                }
+
                 ChoiceGroup::MemSpace => {
                     Box::new(fun.mem_blocks().flat_map(move |block| {
                         let mem_spaces = space.domain().get_mem_space(block.mem_id());
--- a/src/explorer/choice.rs
+++ b/src/explorer/choice.rs
@@ -126,6 +126,35 @@ pub fn list<'a>(
                     }))
                 }
 
+                ChoiceGroup::DimNesting => {
+                    Box::new(fun.dims().enumerate().flat_map(move |(i, lhs)| {
+                        let lhs = lhs.stmt_id();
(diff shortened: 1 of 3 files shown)


src/explorer/config.rs
```diff
@@ -416,9 +422,11 @@ impl<'a> IntoIterator for &'a ChoiceOrdering {
     }
 }
 
-pub(super) const DEFAULT_ORDERING: [ChoiceGroup; 7] = [
+pub(super) const DEFAULT_ORDERING: [ChoiceGroup; 9] = [
```
Elarnon: I don't think we want to change `DEFAULT_ORDERING` until we have more data?  Just to make it harder to confuse ourselves when running a search with default parameters which are not what we thought.

src/explorer/choice.rs
```diff
@@ -125,6 +125,71 @@ pub fn list<'a>(
...
+                                } else {
+                                    None
+                                }
+                            },
+                        )
+                    }))
+                }
+
+                ChoiceGroup::DimFusion => {
+                    // Exposes choices from Order defining whether two
```
Elarnon: I would put those comments on the `ChoiceGroup` enum definition, as a docstring.

src/explorer/choice.rs
```diff
@@ -125,6 +125,71 @@ pub fn list<'a>(
...
+                    Box::new(fun.dims().enumerate().flat_map(move |(i, lhs)| {
+                        let lhs = lhs.stmt_id();
+                        let dims = fun.dims().take(i).map(|x| x.stmt_id());
+
+                        dims.chain(fun.insts().map(|x| x.stmt_id())).flat_map(
+                            move |rhs| {
+                                let available_orders = space.domain().get_order(lhs, rhs);
+                                let nesting_orders = Order::INNER | Order::OUTER;
+
+                                if available_orders.intersects(nesting_orders) {
```
Elarnon: This `if` is not necessary (because if there is no intersection only the last element of the chain will be present).

src/explorer/choice.rs
```diff
@@ -125,6 +125,71 @@ pub fn list<'a>(
...
+                                    let orders = (available_orders & Order::INNER)
+                                        .into_option()
+                                        .into_iter()
+                                        .chain(
+                                            (available_orders & Order::OUTER)
+                                                .into_option()
+                                                .into_iter(),
```
Elarnon: This `.into_iter()` (and the next one) are not necessary due to `.chain` taking an `impl IntoIterator`

src/explorer/config.rs
andidr: Agreed. Will change.

andidr:
bors r=Elarnon

andidr:
bors retry

andidr:
bors retry

andidr:
bors cancel

andidr:
bors r=Elarnon

andidr:
bors retry

andidr:
bors retry

andidr:
bors retry

andidr:
bors retry

andidr:
bors retry

andidr:
bors retry

andidr:
bors retry

PR #275, opened 2019-07-30
Add ability to restart the search

Elarnon:
This adds the ability for the MCTS search algorithm to perform a (soft)
restart.  The soft restart erases information related to the evaluation
statistics (such as average runtime or top-k best runs).

Some information carries over across restarts, most notably:

 - The current cut
 - Nodes which were found dead through backtracking
 - Performance model bounds

The results of constraint propagation are not kept; however, the
structure of the tree ensures that we won't perform contraint
propagation if the result has already failed (or was cut due to the
performance model) in a previous restart.


Elarnon:
bors r=andidr

PR #276, opened 2019-08-02
Add kernels ResNetCell and TransformerCell

andidr:
This series of commits adds two new ML kernels `ResNetCell` and `TransformerCell`. Additionally, for each of these kernels, two further kernels are added that result result from splitting the original kernel in two stages (`ResNetCellTopHalf`, `ResNetCellBottomHalf`, `TransformerCellTopHalf`, and `TransformerCellTopHalf`).

### DIFF
--- a/kernels/src/linalg.rs
+++ b/kernels/src/linalg.rs
@@ -806,3 +806,180 @@ impl<'a, S: Scalar> Kernel<'a> for Fused2MM<'a, S> {
         }
     }
 }
+
+#[derive(Clone, Deserialize, Serialize)]
+pub struct ResNetCellP {
+    pub m: i32,
+    pub n: i32,
+    pub k: i32,
+    pub transpose_a: bool,
+    pub transpose_b: bool,
+    pub transpose_c: bool,
+    pub generic: bool,
+    pub m_tiling: Option<helper::TilingPattern>,
+    pub n_tiling: Option<helper::TilingPattern>,
+    pub k_tiling: Option<helper::TilingPattern>,
+    pub activation_fun: Option<ActivationFunction>,
+}
+
+impl ResNetCellP {
+    pub fn new(m: i32, n: i32, k: i32) -> Self {
+        ResNetCellP {
+            m,
+            n,
+            k,
+            transpose_a: false,
+            transpose_b: false,
+            transpose_c: false,
+            generic: true,
+            m_tiling: None,
+            n_tiling: None,
+            k_tiling: None,
+            activation_fun: None,
+        }
+    }
+
+    pub fn transpose_a(mut self) -> Self {
+        self.transpose_a = true;
+        self
+    }
+
+    pub fn transpose_b(mut self) -> Self {
+        self.transpose_b = true;
+        self
+    }
+
+    pub fn transpose_c(mut self) -> Self {
+        self.transpose_c = true;
+        self
+    }
+
+    pub fn activation_fun<F>(mut self, fun: F) -> Self
+    where
+        F: Into<Option<ActivationFunction>>,
+    {
+        self.activation_fun = fun.into();
+        self
+    }
+
+    /// Inline the sizes in the generated code.
+    pub fn static_sizes(mut self) -> Self {
+        self.generic = false;
+        self
+    }
+}
+
+/// Computes `O = activation(activation(A.B).C) + A`
+pub struct ResNetCell<'a, S: Scalar> {
+    pub params: ResNetCellP,
+    a: Tensor<'a, S>,
+    b: Tensor<'a, S>,
+    c: Tensor<'a, S>,
+    o: Tensor<'a, S>,
+}
+
+impl<'a, S: Scalar> Kernel<'a> for ResNetCell<'a, S> {
+    type Parameters = ResNetCellP;
+    type ExpectedOutput = Array2<S>;
+
+    fn name() -> &'static str {
+        "resnetcell"
+    }
+
+    fn build_signature<AM>(
+        params: ResNetCellP,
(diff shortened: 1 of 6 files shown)


kernels/src/linalg.rs
```diff
@@ -835,3 +808,939 @@ impl<'a, S: Scalar> Kernel<'a> for Fused2MM<'a, S> {
...
+    where
+        F: Into<Option<ActivationFunction>>,
+    {
+        ResNetCellTopHalfP {
+            mm_params: FusedMMP::new(m, n, k).activation_fun(activation_fun),
+        }
+    }
+}
+
+/// Computes `O = activation(A.B)`
```
Elarnon: This is just `FusedMM`; why do we need a separate kernel?

kernels/src/linalg.rs
```diff
@@ -835,3 +808,939 @@ impl<'a, S: Scalar> Kernel<'a> for Fused2MM<'a, S> {
...
+        &self,
+        expected: &Self::ExpectedOutput,
+        context: &dyn device::Context,
+    ) -> Result<(), String> {
+        self.fmmp.check_result(expected, context)
+    }
+}
+
+#[derive(Clone, Deserialize, Serialize)]
+pub struct ResNetCellBottomHalfP {
```
Elarnon: A more explicit name (in NN terminology) could be `FusedDenseBias` -- this is a fused operation performing `activation(a.b) + c`, which corresponds to a fused dense layer (`activation(a.b)`) and bias layer (which is a tensor addition).

kernels/src/linalg.rs
```diff
@@ -835,3 +808,939 @@ impl<'a, S: Scalar> Kernel<'a> for Fused2MM<'a, S> {
...
+        let mut builder = helper::Builder::new(signature, ctx.device());
+
+        let act_ab = self
+            .act_ab
+            .load(vec![n_tiling.clone(), m_tiling.clone()], &mut builder);
+        let c = self.c.load(vec![m_tiling, n_tiling.clone()], &mut builder);
+        let a = self.a.load(vec![n_tiling, k_tiling], &mut builder);
+
+        let act_ab_c = matrix_matrix_multiply(&mut builder, &act_ab, &c);
+        let act_ab_c_pa = matrix_matrix_multiply(&mut builder, &act_ab_c, &a);
```
Elarnon: Why do we have two matrix-matrix multiply here?  Isn't this the code for the full resnet cell?

andidr: This should have been an addition and should be after the activation.

kernels/src/linalg.rs
```diff
@@ -835,3 +808,939 @@ impl<'a, S: Scalar> Kernel<'a> for Fused2MM<'a, S> {
...
+        let qk_scexp = self
+            .qk_scexp
+            .load(vec![m_tiling, r_tiling.clone()], &mut builder);
+
+        let v = self
+            .v
+            .load(vec![r_tiling.clone(), n_tiling.clone()], &mut builder);
+
+        let qk_scexp_div = tensor_elementwise_div(&mut builder, &qk_scexp, &s_exp_op);
+        let o = matrix_matrix_multiply(&mut builder, &qk_scexp_div, &v);
```
Elarnon: FWIW we will replicate the division by the sumexp anyways, so I don't know if we gain that much by using a temp buffer for the exponential here.  Still makes sense to do it?  I think the resnet/convolutions are higher priority anyways

kernels/src/linalg.rs
andidr: I prefer having an explicit `ResNetCellTopHalf`, since there is a `ResNetCellBottomHalf`. We could just type alias `FusedMM`, but this would break the naming scheme for the parameter struct (`ResNetCellTopHalf` would take an instance of `FusedMMP` for its parameters).

kernels/src/linalg.rs
andidr: Not sure either. Let's just keep it for now.

kernels/src/linalg.rs
andidr: I like the more meaningful name. However I'd like to keep a naming scheme that expresses the split of `ResNetCell`. Alternatively, we could use more meaningful names for the split kernels and just add aliases in `cuda_search` of `telamon-cli`. I'm open for suggestions.

kernels/src/linalg.rs
Elarnon: I think there are some benefits to limit the number of kernels that we have, instead of aliasing them (the point is that the kernels we consider are used by a lot of different architectures).

From the pov of a library we wouldn't expose a "ResNetHalf" either, but rather "dense" and "fused_dense_bias" or some such.  So I am very doubtful of the benefits here.

kernels/src/linalg.rs
Elarnon: To me `FusedDenseBias` is much more explicit than `ResNetBottomHalf`, and IMO would be as well for anyone with NN experience (the equivalent operation in tensorflow is called `fused_bias_conv2d` or something similar IIRC).  The point is that this fused layer is used/useful in more than only resnet cells.

andidr: OK, then let's completely eliminate the relationship of the smaller kernels with `TransformerCell` and `ResNetCell` and choose meaningful names for the `*{Top,Bottom}Half` kernels.

- Remove `ResNetCellTopHalf` entirely, since `FusedMM` is already sufficient
- Rename `ResNetBottomHalf` to `FusedDenseBias`
- Rename `TransformerCellBottomHalf` to `ScaledMM`

Any suggestions for `TransformerCellTopHalf`?

PR #278, opened 2019-08-21
Add documentation in codegen::Dimension

Elarnon:
This patch simply adds some documentation to the `codegen::Dimension`
definitions with the result of me trying to understand what is going on
here, which might be useful to future readers of the code.

### DIFF
--- a/src/codegen/dimension.rs
+++ b/src/codegen/dimension.rs
@@ -6,12 +6,30 @@ use std;
 use utils::*;
 
 /// An iteration dimension composed of one or mure fused dimensions.
+///
+/// Note that induction levels are only associated with IR dimensions that are actually used by an
+/// instruction.  For instance, if dimensions `%0` and `%2` are merged in the following code:
+///
+///   @0[%0]: load(a[%0])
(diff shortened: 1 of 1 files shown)


src/codegen/dimension.rs
```diff
@@ -122,10 +140,21 @@ pub fn group_merged_dimensions<'a>(space: &'a SearchSpace) -> Vec<Dimension<'a>>
...
 /// An induction level associated to a dimension.
+///
+/// This represents a variable in a loop that gets incremented at each iteration.
 #[derive(Debug)]
 pub struct InductionLevel<'a> {
+    /// The identifier for the induction variable.  This is shared across the different levels that
+    /// represent a given iteration.
     pub ind_var: ir::IndVarId,
+    /// The increment to use, i.e. the quantity we increment the variable by every iteration.
+    ///
```
andidr: Remove one newline

Elarnon:
What I meant here is: I tried to understand what is going on, and the result (= what I understood) is written done in the documentation.  Agreed it could be more clear; will rephrase.

PR #281, opened 2019-08-21
[explorer] Dump actions.json for best candidates

Elarnon:
Previously we were generating a `best_X.c` and `best_X.cfg` for each
best candidate, which doesn't allow to easily re-use the found
implementation e.g. in the debugger.  The `tlcli rebuild` command was
introduced for this purpose, and can extract a replay file from the
eventlog.

However, evenlogs these days tend to be quite large, and it may not be
desirable to keep them for runs which don't need specific performance
analysis.  Hence, this patch changes the mechanism for dumping best
candidates: now a directory `best_X` is created, with the generated code
in `code.c`; the loop tree in `code.cfg`; and the replay in
`actions.json` (not unlike the format used for errors, which also
contain extra files about the error).

### DIFF
--- a/src/explorer/monitor.rs
+++ b/src/explorer/monitor.rs
@@ -10,6 +10,7 @@ use futures::prelude::*;
 use futures::{executor, future, task, Async};
 use log::warn;
 use serde::{Deserialize, Serialize};
+use std::io::Write;
 use std::sync::{
     self,
     atomic::{AtomicBool, Ordering},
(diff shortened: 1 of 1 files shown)


andidr:
> Previously we were generating a `best_X.c` and `best_X.cfg` for each
> best candidate, which doesn't allow to easily re-use the found
> implementation e.g. in the debugger. The `tlcli rebuild` command was
> introduced for this purpose, and can extract a replay file from the
> eventlog.
> 
> However, evenlogs these days tend to be quite large, and it may not be

even**t**logs

> desirable to keep them for runs which don't need specific performance
> analysis. Hence, this patch changes the mechanism for dumping best
> candidates: now a directory `best_X` is created, with the generated code
> in `code.c`; the loop tree in `code.cfg`; and the replay in
> `actions.json` (not unlike the format used for errors, which also
> contain extra files about the error).

PR #282, opened 2019-08-21
[explorer] Make binary event log optional

Elarnon:
Since we are now dumping the replay files directly for all best
candidates found, the binary event log is only needed for performance
analysis (of the search algorithm, or of Telamon more generally).

This patch hence allows not generating it at all, and makes this the
default configuration.  Users that *do* want the event log should
explicitely set it in the configuration.

### DIFF
--- a/src/explorer/config.rs
+++ b/src/explorer/config.rs
@@ -27,8 +27,9 @@ pub struct Config {
     pub output_dir: String,
     /// Name of the file in wich to store the logs.
     pub log_file: String,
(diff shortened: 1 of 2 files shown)


andidr:
> Since we are now dumping the replay files directly for all best
> candidates found, the binary event log is only needed for performance
> analysis (of the search algorithm, or of Telamon more generally).
> 
> This patch hence allows not generating it at all, and makes this the
> default configuration. Users that _do_ want the event log should
> explicitely set it in the configuration.

explicitly

PR #283, opened 2019-08-21
[cuda] Assume Maxwell and Pascal are dual issue

Elarnon:
The C Programming Guide says that recent (5.x and later)
architectures can issue a single instruction per cycle, but other nvidia
documentation[1] says that dual issue is still possible.  This is
coherent with generated SASS assembly, as well as practical
measurements: the bounds from the performance model are way too high for
some kernels, and changing this setting drops them to reasonable levels.

As noted in the nvidia blog[1], there are apparently more limitations on
those architectures however; such as only being able to issue one load
and one arithmetic operation at the same time (and SASS examination
shows that there might be additional restrictions, e.g. it looks like in
some cases, 128-bit loads can't be dual-issued).  This is not an issue,
since the performance model is optimistic anyways.

Hence, this patch changes the performance model to ignore the
Programming Guide and assume dual-issue for those architectures.

 1: https://devblogs.nvidia.com/5-things-you-should-know-about-new-maxwell-gpu-architecture/

### DIFF
--- a/backend/cuda/src/characterize/gpu.rs
+++ b/backend/cuda/src/characterize/gpu.rs
@@ -267,6 +267,13 @@ fn gpu_rates(gpu: &Gpu, smx_rates: &InstDesc) -> InstDesc {
 /// Returns the number of wrap scheduler in a single SMX and the number of independent
 /// instruction issued per wrap scheduler.
 /// This comes from the "Architecture" subsection for each compute capability on the CPG.
+///
+/// The CPG says that recent (5.x and later) architectures can issue a single instruction per
+/// cycle, but other nvidia documentation [1] says that dual issue is still possible.  This is
+/// coherent with generated SASS assembly, as well as practical measurements -- hence, we ignore
+/// the CPG and assume dual-issue for those architectures.
+///
+/// [1]: https://devblogs.nvidia.com/5-things-you-should-know-about-new-maxwell-gpu-architecture/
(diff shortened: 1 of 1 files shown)


backend/cuda/src/characterize/gpu.rs
```diff
@@ -267,6 +267,13 @@ fn gpu_rates(gpu: &Gpu, smx_rates: &InstDesc) -> InstDesc {
 /// Returns the number of wrap scheduler in a single SMX and the number of independent
 /// instruction issued per wrap scheduler.
 /// This comes from the "Architecture" subsection for each compute capability on the CPG.
+///
+/// The CPG says that recent (5.x and later) architectures can issue a single instruction per
+/// cycle, but other nvidia documentation[1] says that dual issue is still possible.  This is
```
andidr: Just nitpicking: add space before [1]

backend/cuda/src/characterize/gpu.rs
```diff
@@ -267,6 +267,13 @@ fn gpu_rates(gpu: &Gpu, smx_rates: &InstDesc) -> InstDesc {
...
+/// coherent with generated SASS assembly, as well as practical measurements -- hence, we ignore
+/// the CPG and assume dual-issue for those architectures.
+///
+/// 1: https://devblogs.nvidia.com/5-things-you-should-know-about-new-maxwell-gpu-architecture/
```
andidr: Would prefer to either use Markdown notation `[1]: url` or "scientific" reference `[1] url`

PR #284, opened 2019-08-21
[cuda] Correctly handle memory replays for Maxwell and later

Elarnon:
The instruction replay behavior has changed in Maxwell compared to
earlier designs, and is now handled by the individual units, not the
scheduler [1].  As such, for compute capabilities 5 and later, we need
to update the memory model to handle this properly; otherwise, we end up
with a pressure on `issue` that is way too high compared to the reality.

[1]: https://stackoverflow.com/questions/35566178/how-to-explain-instruction-replay-in-cuda


Elarnon:
Superseeded by #307

PR #286, opened 2019-08-21
[cuda] Allow non-coherent loads

Elarnon:
The memory model does not support non-coherent loads; however, they are
mandatory to get a non-negligible performance boost on Maxwell and more
recent architectures.   As an intermediate measure to properly
implementing them in the memory model, this patch allows using them, but
ignores them in the performance model.

### DIFF
--- a/backend/cuda/src/gpu.rs
+++ b/backend/cuda/src/gpu.rs
@@ -241,7 +241,10 @@ impl Gpu {
     /// Returns the description of a load instruction.
     fn load_desc(&self, mem_info: &MemInfo, flags: InstFlag) -> InstDesc {
         // TODO(search_space,model): support CA and NC flags.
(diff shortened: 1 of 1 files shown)


backend/cuda/src/gpu.rs
```diff
@@ -241,7 +241,10 @@ impl Gpu {
     /// Returns the description of a load instruction.
     fn load_desc(&self, mem_info: &MemInfo, flags: InstFlag) -> InstDesc {
         // TODO(search_space,model): support CA and NC flags.
-        assert!(InstFlag::COHERENT.contains(flags));
+        if !InstFlag::COHERENT.contains(flags) {
+            warn!("Non-coherent load.  Bound may be invalid.");
```
andidr: Total nitpicking: two spaces between phrases

PR #290, opened 2019-08-21
[tlcli] Add a `benchmark` command

Elarnon:
This patch adds a `benchmark` command to the telamon CLI, allowing to
re-run benchmarks (both of the generated code, and the reference code)
for a given kernel and replay file after-the-fact.

This will print the same information as in the `benchmark.txt` generated
during the search, namely the average runtime of the benchmarked kernel
and the reference.   The bound for the fixed candidate generated is also
displayed.

A batch mode is also provided, for running the same kernel with
different replays.  In that case, the output (to stdout) is a .csv
containing one line per replay, with the raw measurements.

### DIFF
--- a/telamon-cli/src/bin/tlcli.rs
+++ b/telamon-cli/src/bin/tlcli.rs
@@ -1,16 +1,19 @@
 use std::borrow::Cow;
 use std::collections::HashMap;
-use std::io::{self, Write};
+use std::ffi::{OsStr, OsString};
+use std::fs;
+use std::io::{self, BufRead, BufReader, Write};
 use std::path::PathBuf;
 use std::sync::atomic;
 
+use itertools::*;
 use serde_json;
 use structopt::StructOpt;
 
 use telamon::device;
 use telamon::explorer::{
     self,
-    choice::{ActionEx as Action, Choice},
+    choice::{default_list, ActionEx as Action, Choice},
     config,
     eventlog::EventLog,
     mcts, Candidate,
@@ -18,8 +21,9 @@ use telamon::explorer::{
 use telamon::model::{bound, Bound};
 use telamon::offline_analysis::tree::CandidateTree;
 use telamon::search_space::SearchSpace;
+use telamon_kernels::statistics::estimate_mean;
 
-use telamon_cli::{KernelParam, Platform, ReplayPath};
+use telamon_cli::{Bench, KernelBundle, KernelParam, Platform, ReplayPath};
 
 /// Compute the bound for a given candidate.
 #[derive(StructOpt)]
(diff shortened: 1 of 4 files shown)


telamon-cli/src/bin/tlcli.rs
```diff
@@ -230,6 +234,156 @@ impl Codegen {
...
+
+    fn run(&self, _args: &Opt) -> io::Result<()> {
+        let builder = self.platform.to_builder();
+        let mut context = builder.build_context();
+        let (bundle, context) = context.kernel_bundle(&self.kernel);
+        assert!(bundle.candidates.len() == 1);
+
+        let reference = Bench::default()
+            .warmup(4)
+            .runs(40)
```
andidr: Would suggest to either make these values parametrizeable or make them the default in `Bench::default`

telamon-cli/src/lib.rs
```diff
@@ -227,7 +225,7 @@ mod cuda_reference {
             };
             time_cuda(|| {
                 check_cublas(cublasSgemm_v2(
-                    handle.0, op_b, op_a, n, m, k, &2., b, ldb, a, lda, &3., c, n,
+                    handle.0, op_b, op_a, n, m, k, &1., b, ldb, a, lda, &0., c, n,
```
andidr: Not sure if these changes are supposed to be in this commit.

telamon-cli/src/lib.rs
```diff
@@ -262,8 +260,8 @@ mod cuda_reference {
             let stride_c = (m * n) as libc::c_long;
             time_cuda(|| {
                 check_cublas(cublasSgemmStridedBatched(
-                    handle.0, op_b, op_a, n, m, k, &2., b, ldb, stride_b, a, lda,
-                    stride_a, &3., c, n, stride_c, batch,
+                    handle.0, op_b, op_a, n, m, k, &1., b, ldb, stride_b, a, lda,
+                    stride_a, &0., c, n, stride_c, batch,
                 ));
```
andidr: Same here

telamon-cli/src/lib.rs
Elarnon: Yes they are -- we are now checking that the reference also computes the correct output, and the reference was not computing the correct output (`C = 2 * A * B + 3 * C` instead of `C = A * B`).

telamon-cli/src/lib.rs
Elarnon: See above

telamon-cli/src/bin/tlcli.rs
Elarnon: They actually are the default, but I prefer making this explicit here.  Agree they should be parameters.

Elarnon:
Changed num bench runs to a parameter.

PR #292, opened 2019-08-21
Remove lifetime from codegen::Size

Elarnon:
Since we now use `Arc` for representing parameters, this patch removes
the last use of reference for them in codegen::Size and
codegen::ParamVal.  Removing the lifetimes in those two types simplifies
the lifetime constraints in the code generation phase, and allows to
remove the two remaining uses of `unsafe` in that module that were
(presumably) born out of frustration.

### DIFF
--- a/src/codegen/dimension.rs
+++ b/src/codegen/dimension.rs
@@ -30,7 +30,7 @@ pub struct Dimension<'a> {
     /// The induction levels corresponding to this dimension.
     induction_levels: Vec<InductionLevel<'a>>,
     /// The size of this iteration dimension.
-    size: codegen::Size<'a>,
+    size: codegen::Size,
 }
 
 impl<'a> Dimension<'a> {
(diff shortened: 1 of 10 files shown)


andidr:
> Since we now use `Arc` for representing parameters, this patch removes
> the last use of reference for them in codegen::Size and

`codegen::Size`

> codegen::ParamVal. Removing the lifetimes in those two types simplifies

`codegen::ParamVal`

> the lifetime constraints in the code generation phase, and allows to
> remove the two remaining uses of `unsafe` in that module that were
> (presumably) born out of frustration.

That's speculation ;)

src/codegen/dimension.rs
```diff
@@ -151,9 +151,7 @@ pub struct InductionLevel<'a> {
     ///
     /// Warning: The dimension here corresponds to the IR dimension represented by this level; it
     /// does *not* have the same size as the increment!
-    pub increment: Option<(ir::DimId, codegen::Size<'a>)>,
-    /// The base of the induction, i.e. the value the variable is initialized to before the first
-    /// iteration.
```
andidr: Seems like the comment for `base` got deleted, too

PR #296, opened 2019-08-28
Extract literal parsing to its own trait

Elarnon:
This patch extracts the conversion from `ScalarArgument` to `Operand`
into separate `IntLiteral` and `FloatLiteral` traits, which can be used
to create the literals directly without going through an actual
`Operand`.  This ability is currently not used, but should simplify code
generation since it allows to effortlessly convert Rust literals into
their Telamon representation.

### DIFF
--- a/src/ir/operand.rs
+++ b/src/ir/operand.rs
@@ -1,5 +1,7 @@
 //! Describes the different kinds of operands an instruction can have.
+use std::borrow::Cow;
 use std::fmt;
+use std::sync::Arc;
 
 use self::Operand::*;
 use crate::ir::{self, DimMap, InstId, Instruction, Parameter, Type};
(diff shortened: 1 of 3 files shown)


src/ir/operand.rs
```diff
@@ -9,9 +11,88 @@ use num::bigint::BigInt;
 use num::rational::Ratio;
 use num::traits::{Signed, Zero};
 use serde::{Deserialize, Serialize};
-use std::sync::Arc;
 use utils::unwrap;
 
+/// Trait for representing integer literals which can be used in the IR.
+///
+/// This is meant to be used as a trait bound in various functions accepting literals as arguments,
+/// and allows to write generic function accepting both Rust primitive integer types and IR
```
andidr: generic function**s**

src/ir/operand.rs
```diff
@@ -9,9 +11,88 @@ use num::bigint::BigInt;
...
+impl<'a> IntLiteral<'a> for (&'a BigInt, u16) {
+    fn decompose(self) -> (Cow<'a, BigInt>, u16) {
+        (Cow::Borrowed(self.0), self.1)
+    }
+}
+
+/// Trait for representing floating-point literals which can be used in the IR.
+///
+/// This is meant to be used as a trait bound in various functions accepting literals as arguments,
+/// and allows to write generic function accepting both Rust primitive floating-point types and IR
```
andidr: same here

PR #299, opened 2019-09-02
Use an actual IR for code generation

Elarnon:
This patch introduces an actual low-level IR used for code generation,
as started in #297.  More specifically, it introduces additional
constructs in the `llir` module to represent instructions and operators,
and rewrites the printer code paths to go through these constructs.  In
addition, there are some extra refactorings sneaking in in the x86 and
mppa backends, including the removal of `no_args` templates.

The printers now only have two methods, to display a label and to
display a (predicated) instruction; the printing of both is in practice
done through the `C99Display` (moved to a new `C` backend crate) and
`PTXDisplay` traits.

Additional changes are made to the C-based backends:

 - They now use macros for non-standard operations such as fused
   multiply-adds, high/wide multiplications, or synchronization.  This
   allows decoupling the printing from the implementation details of
   those primitives (e.g. using regular C operators or intrinsics; or
   using a specific threading library in the case of __sync).

 - The way argument passing works has been unified with the PTX
   representation; namely, all arguments must be passed by reference.
   Previously, non-pointer arguments were passed by reference (i.e. a
   pointer to the value) while pointer arguments we passed by value
   (i.e.  the pointer itself); now, pointer arguments are also passed by
   reference (i.e. a pointer to the pointer).

Like #297, this patch should introduce (almost) no functional changes,
and is mostly a large refactoring.  To make sense of the code, the
reader is encouraged to first read through the additions to the `llir`
module; then, the changes in the various printers should be
straightforward translations from the old printing methods to the
`XXXDisplay` implementations.


backend/mppa/src/template/host.c.template
```diff
@@ -9,7 +9,41 @@
...
+
+#define __mul32Hi(a, b) (((int32_t)__mul32Wide(a, b)) >> 32)
+#define __mul32Wide(a, b) ((int64_t)a * (int64_t)b)
+
+#define __mad(a, b, c) (((a) * (b)) + (c))
+#define __mad32Hi(a, b, c) (__mul32Hi(a, b) + c)
+#define __mad32Wide(a, b, c) (__mul32Wide(a, b) + c)
+
+#define __fma(a, b, c) (((a) * (b)) + (c))
+
```
andidr: Not really a fan of the macros (maybe that `telamon_op_max` shouldn't have been a macro in the first place). However, providing inline functions specialized for each type is maybe a bit too verbose. Definitions using `_Generic` aren't probably much shorter either.

PR #301, opened 2019-11-26
Add script expenv-check verifying GPU and CPU settings for experiments

andidr:
The script expenv-check verifies settings for the system's GPUs and CPUs
specified on the command line. These can be:

  - The CPU scaling governor
  - GPU autoboost
  - GPU core frequency
  - GPU memory frequency

By default, settings are only checked and the script fails if the actual
settings and the expected settings do not match. If the --force flag is
specified, the script tries to enfore the specified settings by invoking
appropriate tools (cpufreq-set and nvidia-smi) and only fails if the actual
settings do not match the specified settings in a second check after
modification.

Forcing settings might require certain system settings if expenv-check is
invoked as a user other than root. The installation script expenv-install
provides a setup procedure:

 - Install appropriate systemd services:
   - nvidia-disable-autoboost: Disables GPU autoboost by default
   - nvidia-enable-persistence: Enables persistence mode for GPU settings
   - nvidia-permissions-autoboost: Allows any user to enable/disable autoboost
   - nvidia-permissions-freqs: Allows any user to set GPU core/memory frequencies

 - Install a sudo configuration file (cpufreq-set-non-root), allowing users of
   the group "cpufreq" to invoke cpufreq-set with root privileges and without
   entering a password

 - Add the group "cpufreq" if necessary

 - Install a modprobe configuration file (nvidia-cupti-allow-nonadmin.conf) that
   allows any user to use nvidia CUPTI

The installation is done by simply invoking the installation script as root,
e.g.:

  $ cd tools/expenv
  $ sudo ./expenv-install

A reboot might be necessary after the installation. Users that should be able to
set CPU scaling governors must be added to the group "cpufreq".

### DIFF
--- a/tools/expenv/expenv-check
+++ b/tools/expenv/expenv-check
@@ -1,0 +1,387 @@
+#!/bin/bash
+
+FORCE="false"
+
+CHECK_CPUGOV="false"
+REQ_CPUGOV=""
+
+CHECK_GPUFREQ="false"
+REQ_ACTUAL_GPUFREQ=""
+REQ_SYMBOLIC_GPUFREQ=""
+
+CHECK_GPUMEMFREQ="false"
+REQ_ACTUAL_GPUMEMFREQ=""
+REQ_SYMBOLIC_GPUMEMFREQ=""
+
+CHECK_GPUFREQS="false"
+
+CHECK_GPUAUTOBOOST="false"
+REQ_GPUAUTOBOOST=""
+
+VERBOSE="false"
+
+die() {
+    echo "$1" >&2
+    exit 1
+}
+
+print_help() {
+    echo "Usage: $0 [options]"
+    echo "Check GPU and CPU settings for experiments. By default, settings are"
+    echo "only checked and the script fails if one or more settings do not match"
+    echo "the expected settings. Settings are only adjusted if the --force flag"
+    echo "is specified."
+    echo
+    echo "Options:"
+    echo "  -a, --max-all            Check for maximum performance (equivalent to"
+    echo "                           --cpu-governor=performance --autoboost-enabled"
+    echo "                           --gpu-freq=max --gpu-memfreq=max)"
+    echo "  -c G,--cpu-governor=G    Check if cpu governor is G (e.g., performance)"
+    echo "  -d, --autoboost-disabled Check if GPU autoboost is disabled"
+    echo "  -e, --autoboost-enabled  Check if GPU autoboost is enabled"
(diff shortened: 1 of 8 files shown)


tools/expenv/expenv-check
```diff
@@ -0,0 +1,387 @@
...
+    echo "Options:"
+    echo "  -a, --max-all            Check for maximum performance (equivalent to"
+    echo "                           --cpu-governor=performance --autoboost-enabled"
+    echo "                           --gpu-freq=max --gpu-memfreq=max)"
+    echo "  -c G,--cpu-governor=G    Check if cpu governor is G (e.g., performance)"
+    echo "  -d, --autoboost-disabled Check if GPU autoboost is disabled"
+    echo "  -e, --autoboost-enabled  Check if GPU autoboost is enabled"
+    echo "  -f, --force              Try to enforce desired settings if current settings"
+    echo "                           are different"
+    echo "  -g F, --gpu-freq=F       Check if gpu frequency is G (in MHz) or max"
```
Elarnon: s/G/F

tools/expenv/expenv-check
```diff
@@ -0,0 +1,387 @@
...
+    echo "  -h, --help               Show this help"
+    echo "  -m F, --gpu-memfreq=F    Check if gpu memory frequency is G (in MHz) or max"
```
Elarnon: s/G/F

tools/expenv/expenv-install
```diff
@@ -0,0 +1,113 @@
...
+	die "Could not install $TGT"
+    echo "done."
+}
+
+MODPROBED_DIR="/etc/modprobe.d"
+SYSTEMD_DIR="/etc/systemd/system"
+BIN_DIR="/usr/bin"
+SUDOERSD_DIR="/etc/sudoers.d"
+
+cd "`dirname $0`"
```
Elarnon: Move below `SERVICES=` definition to keep all variable definitions together?

tools/expenv/expenv-check
```diff
@@ -0,0 +1,387 @@
...
+}
+
+gpu_mhz_val() {
+    local GPUIDX="$1"
+    local PARAM="$2"
+
+    nvidia-smi -i $GPUIDX \
+	       --query-gpu=$2 \
+	       --format=csv,noheader | \
+	grep -o "[0-9]\+\sMHz" | grep -o "[0-9]\+"
```
Elarnon: `\s\+` is more robust

PR #303, opened 2019-11-26
[nvmon] Flush on SIGTERM

Elarnon:
The benchmark scripts use nvmon to monitor the frequency during telamon
runs.  nvmon is killed at the end of the run, but currently does not
flush its output -- which means that when stdout is redirected to a
file, we lose the very data we wanted to capture.

This patch adds a SIGTERM handler to nvmon to flush `stdout` before
exiting.

### DIFF
--- a/tools/nvmon/nvmon.c
+++ b/tools/nvmon/nvmon.c
@@ -5,11 +5,13 @@
  */
 
 #include <assert.h>
-#include <nvml.h>
+#include <signal.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
 #include <unistd.h>
+
+#include <nvml.h>
 
 #define NVML_CHECK(f)                                                          \
   {                                                                            \
@@ -19,6 +21,11 @@
       exit(1);                                                                 \
     }                                                                          \
   }
+
+static void on_sigterm(int signum) {
+  fflush(stdout);
+  exit(1);
+}
 
 struct samplesBuffer {
   nvmlSamplingType_t type;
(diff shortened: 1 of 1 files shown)


tools/nvmon/nvmon.c
```diff
@@ -5,12 +5,14 @@
...
 #include <assert.h>
-#include <nvml.h>
+#include <signal.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
 #include <unistd.h>
 
+#include <nvml.h>
+
```
andidr: Is there a particular reason for moving this include down or was this change just committed by accident?

tools/nvmon/nvmon.c
```diff
@@ -403,6 +410,16 @@ void print_samples(void) {
 
 int main(int argc, char **argv) {
   if (argc == 1) {
+    struct sigaction sa;
+    sa.sa_handler = on_sigterm;
+    sigemptyset(&sa.sa_mask);
+    sa.sa_flags = 0;
+
+    if (sigaction(SIGTERM, &sa, NULL) == -1) {
+      fprintf(stderr, "Failed to register SIGTERM handler");
```
andidr: Since no string formatting takes place, `fputs` should be sufficient

tools/nvmon/nvmon.c
Elarnon: It was moved down to separate system includes from CUDA includes.

tools/nvmon/nvmon.c
Elarnon: I prefer using `fprintf` everywhere when performance is not critical for consistency and to make it easier to add string formatting if wanted.  I'm fine with changing it to `fputs` however.

andidr: No strong feelings about this, but `fprintf` looks weird without formatting.

PR #305, opened 2019-11-26
Allow exact tiling factor

Elarnon:
Currently we only allow tiling factors which are strict divisors of the
dimension size.  There are no real resons not to also allow exact
divisors (e.g. tiling a dimension of size 32 by a factor 32 -- this is
still useful because it transforms a dimension of unknown size into a
dimension of known size), so this patch allows it.


andidr:
> Currently we only allow tiling factors which are strict divisors of the
> dimension size. There are no real resons not to also allow exact

reasons

> divisors (e.g. tiling a dimension of size 32 by a factor 32 -- this is
> still useful because it transforms a dimension of unknown size into a
> dimension of known size), so this patch allows it.

PR #306, opened 2019-11-26
Catch panics in CUDA compilation

Elarnon:
We currently allow errors in kernel execution (e.g. if there is a crash
during execution, or results are incorrect) but in case there is an
error during the IR -> PTX compilation process we immediately abort the
search.

This makes experimentations hard, because there are times where those
compilation failures are due to corner cases which are expensive to
handle and shouldn't prevent us from exploring the other interesting
cases.  One example for which this happens currently (and will be fixed
separately) is when using dimension sizes which have distinct prime
factors, we can end up with a invalid tiling factor -- for instance we
could try to tile `15` by `9`.  This causes assertion failures in the
IR -> PTX compiler, which in turn stop the search process.

The "proper" fix for this would be to update the code and use `Results`
in the appropriate places, but that requires more manpower than
available right now.  Instead we use `std::panic::catch_unwind` as a
best-effort recovery mechanism.

### DIFF
--- a/src/device/context.rs
+++ b/src/device/context.rs
@@ -71,9 +71,9 @@ pub trait Context: Sync {
         for p in size.dividend() {
             dividend *= unwrap!(self.param_as_size(&p.name));
         }
-        let (result, remider) = num::integer::div_rem(dividend, size.divisor());
+        let (result, remainder) = num::integer::div_rem(dividend, size.divisor());
         assert_eq!(
-            remider, 0,
+            remainder, 17,
             "invalid size: {:?} (dividend = {})",
             size, dividend
         );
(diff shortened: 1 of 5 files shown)


src/device/context.rs
```diff
@@ -71,9 +71,9 @@ pub trait Context: Sync {
         for p in size.dividend() {
             dividend *= unwrap!(self.param_as_size(&p.name));
         }
-        let (result, remider) = num::integer::div_rem(dividend, size.divisor());
+        let (result, reminder) = num::integer::div_rem(dividend, size.divisor());
```
andidr: remainder

PR #312, opened 2019-11-26
Make nvmon compile again

Elarnon:
This patch also changes a bit the error infrastructure by introducing a
rust-like `panic` macro.  It also ensures successful return on `SIGTERM`
since it is the expected mode of operation for nvmon to be killed.

### DIFF
--- a/tools/nvmon/nvmon.c
+++ b/tools/nvmon/nvmon.c
@@ -14,17 +14,28 @@
 #include <nvml.h>
 
 #define NVML_CHECK(f)                                                          \
-  {                                                                            \
+  do {                                                                         \
     nvmlReturn_t __status = (f);                                               \
     if (__status != NVML_SUCCESS) {                                            \
-      fprintf(stderr, "%s:%d:FAILURE: %d", __FILE__, __LINE__, __status);      \
-      exit(1);                                                                 \
+      panic("NVML error %d", __status);                                        \
     }                                                                          \
-  }
+  } while (0)
+
+#define eprintf(...) fprintf(stderr, __VA_ARGS__)
+
+#define eputs(x) fputs(x, stderr)
+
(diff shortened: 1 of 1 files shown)


tools/nvmon/nvmon.c
```diff
@@ -17,14 +17,25 @@
...
-      fprintf(stderr, "%s:%d:FAILURE: %d", __FILE__, __LINE__, __status);      \
-      exit(1);                                                                 \
+      panic("NVML error %d", __status);                                        \
     }                                                                          \
   }
 
+#define eprintf(...) fprintf(stderr, __VA_ARGS__)
+
+#define eputs(x) fputs(x, stderr)
+
```
andidr: Defining macros for a single point of invocation right below looks is a bit of overkill

tools/nvmon/nvmon.c
```diff
@@ -17,14 +17,25 @@
...
+#define panic(...)                                                             \
+  {                                                                            \
+    eprintf("%s:%d:", __FILE__, __LINE__);                                     \
+    eprintf(__VA_ARGS__);                                                      \
+    eputs("\n");                                                               \
+    exit(1);                                                                   \
+  }
+
```
andidr: The idiomatic way in C to define macros with multiple statements is to wrap this in a `do { ...} while(0)`. This allows the macro to be treated as a single statement. Wrapping in a block is not equivalent, e.g., when using this as the body in an if statement:
```C
/* Does not compile: error: ‘else’ without a previous ‘if’ */
if(...)
    panic("foo");
else
    ...
```

tools/nvmon/nvmon.c
Elarnon: It future-proofs further usages I guess.

PR #314, opened 2020-07-02
Updated paper URL in README.md.

samkaufman:
Suggested fix to broken link.


Elarnon:
Indeed, that link died some time ago, and your fix is correct.  Thank you!

